PR #42, opened 2021-07-06
Added not blank rule and also added the test case for the same. Also …

AmandeepSinghCS:
Added not blank rule and also added the test case for the same. Also updated the README.md


tests/test_templates.py
```diff
@@ -106,6 +106,13 @@ def test_rule_bindings_collection_team_3(self):
             Path("tests/resources/configs/rule_bindings/team-3-rule-bindings.yml")
         )
 
+    @pytest.fixture
+    def test_rule_bindings_collection_team_4(self):
+        """ """
+        return lib.load_rule_bindings_config(
+            Path("tests/resources/configs/rule_bindings/team-4-rule-bindings.yml")
+        )
+
```
thinhha, line 115: is this fixture being used anywhere?

clouddq/classes/rule_type.py
```diff
@@ -26,6 +26,7 @@
 FORBIDDEN_SQL = re.compile(r"[;\-#|\\/]")
 NOT_NULL_SQL = Template("$column IS NOT NULL")
 REGEX_SQL = Template("REGEXP_CONTAINS( CAST( $column  AS STRING), '$pattern' )")
+NOT_BLANK_SQL = Template("$column != ''")
```
thinhha, line 29: what do you think about using `TRIM(col) != ''` to account for blanks with multiple spaces?

thinhha (changes requested):
Thanks Amandeep. Please see a feel minor questions/suggestions.

tests/test_templates.py
thinhha: I would also suggest adding the not blank rule to an existing test fixture instead of a new one, e.g.

`
rule_bindings:
  T2_DQ_1_EMAIL:
    entity_id: TEST_TABLE
    column_id: VALUE
    row_filter_id: DATA_TYPE_EMAIL
    rule_ids:
      - NOT_NULL_SIMPLE
      - REGEX_VALID_EMAIL
      - CUSTOM_SQL_LENGTH_LE_30
      - NOT_BLANK
    metadata:
      brand: one
`

You will most likely need to edit other test resources e.g. the expected json and generated sql.

AmandeepSinghCS: Unit Tested it separately as a single data quality rule.

clouddq/classes/rule_type.py
AmandeepSinghCS: This makes sense, I will make the required changes and test it.

AmandeepSinghCS:
@googlebot I fixed it..

PR #45, opened 2021-07-12
Added Not Blank Rule

AmandeepSinghCS:


tests/resources/test_render_run_dq_main_sql_expected.sql
```diff
@@ -84,6 +84,26 @@ SELECT
...
+    '<your_gcp_project_id>.dq_test.contact_details' AS table_id,
+    'value' AS column_id,
+    value AS column_value,
+    num_rows_validated AS num_rows_validated,
+    CASE
+
+      WHEN value IS NULL THEN NULL
+      WHEN TRIM(value) != '' THEN TRUE
+
+    ELSE
```
thinhha, line 101: FYI @AmandeepSinghCS I've updated the for formatting in this test to be more aligned with the other rules output. The test implicitly strips spaces and empty lines from both test and expected outputs so all tests are passing still but I still think it is still worth aligning on formatting.

thinhha (approved):

thinhha:
LGTM thanks @AmandeepSinghCS !

PR #46, opened 2021-08-10
make duplicate rule return 0 rows and full count

thinhha:
This fixes two bugs with the existing duplicates check rule:

- previously the duplicates rule returned nothing so it did not update the failure count to 0 once the duplicates are removed, with this change it returns 0 as the count
- also it will return the total number of duplicated records not just the distinct count of duplicated records

### DIFF
--- a/configs/rules/complex-rules.yml
+++ b/configs/rules/complex-rules.yml
@@ -20,8 +20,13 @@ rules:
       custom_sql_arguments:
         - column_names
       custom_sql_statement: |-
-        select
-          $column_names
-        from data
-        group by $column_names
-        having count(*) > 1
+        select a.*, duplicates.*
+        from data a
+        inner join (
+          select
+            $column_names
+          from data
+          group by $column_names
+          having count(*) > 1
+        ) duplicates
+        using ($column_names)
--- a/clouddq/templates/dbt/macros/macros.sql
+++ b/clouddq/templates/dbt/macros/macros.sql
@@ -43,11 +43,11 @@
     '{{ rule_id }}' AS rule_id,
     '{{ fully_qualified_table_name }}' AS table_id,
     '{{ column_name }}' AS column_id,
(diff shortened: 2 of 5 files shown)


configs/rules/complex-rules.yml
```diff
@@ -20,8 +20,13 @@ rules:
       custom_sql_arguments:
         - column_names
       custom_sql_statement: |-
-        select
-          $column_names
-        from data
-        group by $column_names
-        having count(*) > 1
+        select a.*, duplicates.*
```
andrew-ii, line 23: General comment for later consideration (not blocking the overall PR approval) - would be good to review the interface with `custom_sql_statement` queries. Right now it seems to be arbitrary, while I believe it needs to be a set interface to allow managing custom queries under a general framework. Again, not a blocker for this specific change

andrew-ii (approved):

configs/rules/complex-rules.yml
thinhha: Thanks Andrew. Currently each custom_sql_statement string is substituted into this macro template: https://github.com/GoogleCloudPlatform/cloud-data-quality/blob/main/clouddq/templates/dbt/macros/macros.sql#L39

Apart from the fact that you can only run a single select query, it is pretty much arbitrary at the moment.

thinhha: I considered updating the macro/rule type itself but it wasn't necessary to get the target outcome for this rule. In theory, once duplicates works properly this rule should generalize to other rule types.

One downside of this approach is that it is not currently possible to retrieve only the failing records. I decided to implement that as part of a separate scope to materialize all failing records not just the complex rules, as that feature will require a few other things to change as well.

thinhha: any further thoughts let me know otherwise I'll merge and release a patch for the current release.

AmandeepSinghCS (approved):

PR #48, opened 2021-09-20
Hide dbt

thinhha:


clouddq/runners/dbt/dbt_runner.py
```diff
@@ -0,0 +1,257 @@
...
+        self.dbt_rule_binding_views_path.mkdir(parents=True, exist_ok=True)
+        logger.info(
+            "Writing generated sql to %s/",
+            self.dbt_rule_binding_views_path.absolute(),
+        )
+
+    def __write_connection_configs_to_dbt_profiles_yml_file(self) -> None:
+        template = load_jinja_template(
+            template="profiles.yml",
+            template_parent=DBT_TEMPLATED_FILE_LOCATIONS["profiles.yml"],
```
thinhha, line 248: update load_jinja_template call to use new signature

clouddq/classes/bigquery_client.py
```diff
@@ -0,0 +1,177 @@
...
+from google.cloud import bigquery
+from google.oauth2 import id_token
+from google.oauth2 import service_account
+
+
+logger = logging.getLogger(__name__)
+
+TARGET_SCOPES = [
+    "https://www.googleapis.com/auth/cloud-platform",
+    "https://www.googleapis.com/auth/userinfo.email",
```
AmandeepSinghCS, line 39: is the user info scope required?

clouddq/classes/bigquery_client.py
```diff
@@ -0,0 +1,177 @@
...
+            self.__project_id = self.__credentials._quota_project_id
+        else:
+            logger.warn(
+                "Could not retrieve project_id from ADC credentials.", exc_info=True
+            )
+
+    def get_connection(self, new: bool = False) -> bigquery.client.Client:
+        """Creates return new Singleton database connection"""
+        if self.__client is None or new:
+            job_config = bigquery.QueryJobConfig(dry_run=True, use_query_cache=False)
```
AmandeepSinghCS, line 133: is dry run required? Also, can we make use_query_cache to True, so that it uses results from cache for same queries and it will be flushed once the table is modified.

clouddq/runners/dbt/dbt_runner.py
```diff
@@ -0,0 +1,233 @@
...
+            logger.debug(
+                f"Cannot find `dbt_project.yml` in path: {dbt_project_path} \n"
+                f"Writing templated file to: {dbt_project_path}/dbt_project.yml"
+            )
+            write_templated_file_to_path(dbt_project_path, DBT_TEMPLATED_FILE_LOCATIONS)
+        logger.debug(f"Using 'dbt_project_path': {dbt_project_path}")
+
+    def __prepare_dbt_main_path(self) -> None:
+        assert self.dbt_path.is_dir()
+        dbt_main_path = self.dbt_path / "models" / "data_quality_engine"
```
AmandeepSinghCS, line 212: Can we create constants for static strings?

clouddq/runners/dbt/dbt_runner.py
```diff
@@ -0,0 +1,233 @@
...
+        dbt_main_path.mkdir(parents=True, exist_ok=True)
+        if not dbt_main_path.joinpath("main.sql").is_file():
+            write_templated_file_to_path(
+                dbt_main_path.joinpath("main.sql"), DBT_TEMPLATED_FILE_LOCATIONS
```
AmandeepSinghCS, line 216: constants for file names

clouddq/classes/bigquery_client.py
```diff
@@ -0,0 +1,177 @@
...
+            source_credentials = credentials
+        # Use service account json key if provided
+        elif gcp_service_account_key_path:
+            source_credentials = service_account.Credentials.from_service_account_file(
+                filename=gcp_service_account_key_path, scopes=TARGET_SCOPES
+            )
+        # Otherwise, use Application Default Credentials
+        else:
+            source_credentials, _ = google.auth.default(
+                scopes=TARGET_SCOPES, quota_project_id=project_id
```
AmandeepSinghCS, line 80: if project id is not provided , it should use current project as default or no?

thinhha: I don't think the quota_project_id is automatically set to the project_id

clouddq/runners/dbt/dbt_runner.py
thinhha: alas Python doesn't have constants :( declaring these variables at global scopes means they could be changed at runtime which we do not want

clouddq/runners/dbt/dbt_runner.py
thinhha: alas Python doesn't have constants :( declaring these variables at global scopes means they could be changed at runtime which we do not want

clouddq/classes/bigquery_client.py
thinhha: you are right this is not needed. I will remove it

clouddq/classes/bigquery_client.py
thinhha: Hmm this was originally meant for the dry-run client. Maybe we can rename this to "get_dry_run_connection" and have a separate method called "get_connection" for non-dry-run queries? Let me make that update.

thinhha: I ended up removing the dry-run configs from `get_connection` and instead moving the dry-run job config to the method `check_query_dry_run` directly

clouddq/runners/dbt/dbt_runner.py
AmandeepSinghCS: can we create a separate constants.py file and import these required variables from the same? we can discuss.

AmandeepSinghCS (approved):
Looks good to me.

PR #58, opened 2021-10-05
Extract the correct dataset

charleskubicek:
The previous code only extracted the project ID so a dataset with the name $project_id:$project_id was being asserted for. This failed the entire process.


thinhha:
I'll also fix the GHA pipeline as it doesn't appear to have anything to do with your PR

thinhha:
The tests are failing as it's using the old gha script. I'll merge this for now and will fix other issues.

PR #63, opened 2021-10-07
Target table

AmandeepSinghCS:
Added DQ Summary target table functionality and updated BQ client.


clouddq/classes/bigquery_client.py
```diff
@@ -202,3 +202,28 @@ def check_query_dry_run(self, query_string: str) -> None:
...
+
+    def is_table_exists(self, table: str) -> bool:
+        try:
+            client = self.get_connection()
+            client.get_table(table)
+            return True
+        except NotFound:
+            return False
+
+    def execute_query(self, query_string: str, job_config) -> bigquery.job.QueryJob:
```
thinhha, line 214: Please add type annotation for job_config, also could we make this an optional argument and provide a default job config if the caller doesn't specify a job_config?

clouddq/classes/bigquery_client.py
```diff
@@ -202,3 +202,28 @@ def check_query_dry_run(self, query_string: str) -> None:
...
+            result of the sql execution is returned
+        """
+        try:
+            client = self.get_connection()
+            logger.debug("Query string is \n %s", query_string)
+            query_job = client.query(query_string, job_config=job_config).result()
+            return query_job
+
+        except Exception as e:
+            logger.error(f"Query Execution failed with error {e}\n", exc_info=True)
```
thinhha, line 229: I think we can just raise the exception here and let the caller handle the error, instead of hiding the error behind a log statement here.

thinhha: in other words the try except clause may not be needed, unless you are going to specifically do something with the caught error

clouddq/classes/bigquery_client.py
```diff
@@ -202,3 +202,28 @@ def check_query_dry_run(self, query_string: str) -> None:
...
+        The method is used to execute the sql query
+        Parameters:
+        query_string (str) : sql query to be executed
+        Returns:
+            result of the sql execution is returned
+        """
+        try:
+            client = self.get_connection()
+            logger.debug("Query string is \n %s", query_string)
+            query_job = client.query(query_string, job_config=job_config).result()
```
thinhha, line 225: can we default the 'job_id_prefix' to 'clouddq-'?

clouddq/classes/dq_target_table_utils.py
```diff
@@ -0,0 +1,106 @@
...
+            write_disposition="WRITE_APPEND",
+        )
+
+        query_string = f"""SELECT * FROM `{dq_summary_table_name}` 
+         WHERE invocation_id='{invocation_id}' 
+         and DATE(execution_ts)='{partition_date}'"""
+
+        # Start the query, passing in the extra configuration.
+        # Make an API request and wait for the job to complete
+        bigquery_client.execute_query(query_string=query_string, job_config=job_config)
```
thinhha, line 48: is this a synchronous call i.e. the execute_query blocks until the query completes? or is it asynchronous and returns a future?

clouddq/classes/bigquery_client.py
thinhha: please also set useLegacySql to False

clouddq/classes/dq_target_table_utils.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def write_to_target_bq_table(
+        self,
+        partition_date: date,
+        target_bigquery_summary_table: str,
+        dq_summary_table_name: str,
+    ):
+        try:
+
+            load_target_table_from_bigquery(
```
thinhha, line 93: is the only reason why `TargetTable.write_to_target_bq_table()` is a one-to-one map to `load_target_table_from_bigquery()` because you expect `TargetTable.write_to_target_bq_table()` to be able to handle other source types in the future?

clouddq/classes/dq_target_table_utils.py
```diff
@@ -0,0 +1,106 @@
...
+                bigquery_client=self.bigquery_client,
+                invocation_id=self.invocation_id,
+                partition_date=partition_date,
+                target_bigquery_summary_table=target_bigquery_summary_table,
+                dq_summary_table_name=dq_summary_table_name,
+            )
+
+        except Exception as e:
```
thinhha, line 101: I suggest bubbling this error upstream by raising it instead of hiding it here

clouddq/main.py
```diff
@@ -132,6 +135,10 @@ def get_logger():
     "rule_binding_config_path",
     type=click.Path(exists=True),
 )
+@click.option(
+    "--target_bigquery_summary_table",
+    help="",
```
thinhha, line 140: please add proper help text and documentation

clouddq/main.py
```diff
@@ -272,25 +279,21 @@ def main(  # noqa: C901
     metadata: Optional[str],
     dry_run: bool,
     progress_watermark: bool,
+    target_bigquery_summary_table: str,
```
thinhha, line 282: this is an optional argument

thinhha: we should also assert that this summary table is in the same region as --gcp_region_id. I don't see the assert_dataset_is_in_region() method call in this main class...does your branch have the latest commit from the main branch?

clouddq/main.py
```diff
@@ -438,6 +439,20 @@ def main(  # noqa: C901
...
+                partition_date = date.today()
+                logger.info("Partition date is %s", partition_date)
+                target_table = TargetTable(invocation_id, bigquery_client)
+                target_table.write_to_target_bq_table(
+                    partition_date,
+                    target_bigquery_summary_table,
+                    dq_summary_table_name
+                )
+            else:
+                raise RuntimeError("Target bigquery summary table is not provided or empty")
```
thinhha, line 455: why are we raising an error here? I thought the target table is optional? We should also not log "Job completed successfully." until this step is complete.

clouddq/main.py
```diff
@@ -438,6 +439,20 @@ def main(  # noqa: C901
         )
         if job_status == JobStatus.SUCCESS:
             logger.info("Job completed successfully.")
+
+            if not dry_run and target_bigquery_summary_table:
+                invocation_id = get_dbt_invocation_id(dbt_path)
```
thinhha, line 444: I suggest moving get_dbt_invocation_id() into the runners.dbt. path somehow, Maybe part of DbtRunner

clouddq/main.py
```diff
@@ -438,6 +439,20 @@ def main(  # noqa: C901
...
+                logger.info("Invocation id is %s ", invocation_id)
```
thinhha, line 445: We should clarify this is dbt Invocation ID, also please use f strings

clouddq/main.py
```diff
@@ -438,6 +439,20 @@ def main(  # noqa: C901
...
+                partition_date = date.today()
+                logger.info("Partition date is %s", partition_date)
```
thinhha, line 447: please use fstring and clarify what this date is used for in the log?

clouddq/runners/dbt/dbt_utils.py
```diff
@@ -149,3 +149,13 @@ def get_bigquery_dq_summary_table_name(
         dbt_dataset = extract_dbt_env_var(dbt_dataset)
     dq_summary_table_name = f"{dbt_project}.{dbt_dataset}.{table_name}"
     return dq_summary_table_name
+
+
+def get_dbt_invocation_id(dbt_path: Path) -> str:
+
+    logger.debug("dbt path is %s", dbt_path)
+    # JSON file
+    with open(f"{dbt_path}/target/manifest.json") as f:
```
thinhha, line 158: please use Pathlib.path and use joinpath instead of hardcoding '/'

clouddq/runners/dbt/dbt_utils.py
```diff
@@ -149,3 +149,13 @@ def get_bigquery_dq_summary_table_name(
...
+        # Reading from file
+        data = json.loads(f.read())
+        return data["metadata"]["invocation_id"]
```
thinhha, line 161: we should log a useful error message if invocation_id is not found (Dbt hasn't run yet)

thinhha (changes requested):
Thanks Amandeep - few feedback points

clouddq/classes/bigquery_client.py
AmandeepSinghCS: Did this only for logging the error. But you are right, this try except clause can be removed as we aren't doing anything specific with the error.

clouddq/classes/dq_target_table_utils.py
AmandeepSinghCS: yes, I made it synchronous by chaining a .result() to client.query() so that it will wait till the BQ job is complete.

clouddq/classes/dq_target_table_utils.py
AmandeepSinghCS: yes exactly

clouddq/main.py
AmandeepSinghCS: updated the log message.

clouddq/runners/dbt/dbt_utils.py
AmandeepSinghCS: Logged the error message, with the error information

clouddq/classes/dq_target_table_utils.py
thinhha: Thanks for confirming Amandeep

PR #71, opened 2021-11-03
make dataproc pyspark job bubble up run failures

thinhha:
Dataproc job current report job status as successful even when the job failed. This is because the subprocess run is not capturing and bubbling up errors in the pyspark driver.

Additionally, the bazel test shell script currently requires GOOGLE_APPLICATION_CREDENTIALS even if it is not set. This change allows the environment variable to be passed in if it exists, otherwise it passes in an empty string if the environment variable does not exists.


pbalm (approved):
LGTM.
Is it possible to add a test that calls this script with illegal arguments, causing the subprocess to fail, and verify that the script exits with non-zero return code?

thinhha:
Thanks Paul - We will create a ticket to add automated integration testing for the Dataproc PySpark job.

PR #75, opened 2021-11-05
Update pyspark driver

thinhha:
1) we will move the clouddq_pyspark_driver.py from scripts/dataproc to clouddq/integration/, 
2) clouddq_pyspark_driver.py will support configs in yaml/yml format as well as zip files with arbitrary content (the zip doesn't need to only contain one directly called 'configs' as it is today) 
3) we will remove scripts/dataproc/* as the original script was a PoC and we do not plan to maintain it since we will replace it with another guide on on deploying CloudDQ with Composer+Dataproc Workflows

Integration tests for the PySpark coming as part of: https://github.com/GoogleCloudPlatform/cloud-data-quality/pull/76/files#diff-b8339c817b25b3ccdd71b6c4b83b1574a78f4d772f58e89fc77f89cd43a24ed6R102


jayBana (approved):
All looks good to me.

PR #77, opened 2021-11-05
Adding usage guide and scripts for CloudDQ with Dataproc Workflow using Composer

jayBana:
Added usage guide and scripts for documenting an example for running the CloudDQ CLI on a schedule with Dataproc Workflow using Cloud Composer. The scripts include a helper script for uploading CloudDQ artifacts to Cloud Storage, an example Composer DAG Python file and a Dataproc Workflow template YAML.


thinhha:
LGTM - thanks Janos!

PR #78, opened 2021-11-06
Fix for go lang env missing after running install_development_dependencies

ihaffa-google:
Currently after running "install_development_dependencies.sh" go is unaccessible.

Fix includes creating a common "env.sh" to store any commonly use enviroment variables. install_development_dependencies.sh will call the env.sh to set the go enviroment.

### DIFF
--- a/scripts/env.sh
+++ b/scripts/env.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


thinhha:
Thanks for the contributions Ihaffa!

I wonder if this issue would have been fixed with https://github.com/GoogleCloudPlatform/cloud-data-quality/pull/68?

We currently reference another script to set up environment variables extensively. I'm hesitant to add another one as it can cause confusion.
https://github.com/GoogleCloudPlatform/cloud-data-quality/blob/main/set_environment_variables.sh

ihaffa-google:
Thank you for the comment! I believe you are right, let me close this. 

Note: source should work (#68) no fix is needed

PR #80, opened 2021-11-09
Relax YAML spec

pbalm:
Permit duplicate keys in YAML spec if definitions are identical.

Note that "identical" is strict: Even if the config may be case insensitive, the comparison of two configs is based on equality of the strings.

### DIFF
--- a/tests/unit/test_lib.py
+++ b/tests/unit/test_lib.py
@@ -88,7 +88,7 @@ class TestLib:
 
 
     # Load a config directory containing two different configs defining the same key
-    def test_load_configs_identical(self, temp_configs_dir):
+    def test_load_configs_different(self, temp_configs_dir):
         try:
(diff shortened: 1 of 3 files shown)


tests/unit/test_lib.py
```diff
@@ -0,0 +1,128 @@
...
+
+            loaded_config = lib.load_configs(temp_dir, DqConfigType.ENTITIES)
+
+            assert list(loaded_config.keys()) == ['TEST_TABLE']
+        finally:
+            shutil.rmtree(temp_dir)
+
+
+    # Load a config directory containing two different configs defining the same key
+    def test_load_configs_identical(self, temp_configs_dir):
```
thinhha, line 91: the two tests have the same name. shouldn't Python complain about this?

thinhha (approved):
LGTM - thanks!

PR #85, opened 2021-11-18
Adding last_modified col to dq_summary

pbalm:
The last_modified column in the `dq_summary` table indicates the time the table on which the validation was run was last updated, at the time it was validated.

### DIFF
--- a/tests/integration/test_cli_integration.py
+++ b/tests/integration/test_cli_integration.py
@@ -22,6 +22,7 @@ import click.testing
 import pytest
 
 from clouddq.main import main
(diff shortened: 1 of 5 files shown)


tests/integration/test_cli_integration.py
```diff
@@ -79,6 +80,59 @@ def test_cli_dry_run_oauth_configs(
         print(result.output)
         assert result.exit_code == 0
 
+    def test_cli_oauth_configs(
```
thinhha, line 83: could you use a different test for checking the last modified column change?

thinhha (approved):
LGTM - thanks

PR #92, opened 2021-11-23
Log dq_summary table to stdout

pbalm:
This allows collection of the result using Cloud Logging or other means, which in turn enables centralized reporting on the results.


thinhha:
Thanks for this Paul. Would you mind adding a flag to enable this feature?

clouddq/log.py
```diff
@@ -0,0 +1,96 @@
...
+            },
+            cls=JsonEncoderDatetime,
+        )
+        return super().format(record)
+
+
+def get_json_logger():
+    json_logger = logging.getLogger("clouddq-json-logger")
+    json_logger.setLevel(LOG_LEVEL)
+    logging_stream_handler = logging.StreamHandler(sys.stderr)
```
thinhha, line 81: we should change this to sys.stdout

clouddq/integration/bigquery/dq_target_table_utils.py
```diff
@@ -65,7 +78,9 @@ def load_target_table_from_bigquery(
         SELECT * from `{dq_summary_table_name}`
         WHERE invocation_id='{invocation_id}'
         AND DATE(execution_ts)='{partition_date}'"""
-        bigquery_client.execute_query(query_string=query_string)
+        summary_data = bigquery_client.execute_query(query_string=query_string).result()
```
thinhha, line 81: are you sure executing a DML CREATE TABLE AS SELECT returns RowIterator?

thinhha (commented):
one minor suggestion but otherwise LGTM - thanks Paul!

clouddq/integration/bigquery/dq_target_table_utils.py
pbalm: It does but it's an EmptyRowIterator :-/ I have refactored a bit to run first the SELECT and get the data, then CREATE TABLE to create the target summary table.

PR #99, opened 2021-11-29
fix get_json_logger logic

thinhha:


pbalm (approved):
Tested by building executable and checking behaviour, works.

pbalm:
For the record: I think the difference between the code that we had and the code after this merge is that `hasHandlers()` returns to if this logger has zero handlers, but one of the parents in the logger hierarchy has handlers. Whereas the current code checks only this logger.

pbalm:
https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L1662

PR #115, opened 2021-12-09
added test for dq rules

AmandeepSinghCS:


tests/integration/test_dq_rules.py
```diff
@@ -0,0 +1,153 @@
...
+        target_bq_result_table_name,
+        tmp_path,
+        client,
+        gcp_impersonation_credentials,
+        gcp_sa_key,
+        create_expected_results_table,
+    ):
+
+        try:
+            temp_dir = Path(tmp_path).joinpath("clouddq_test_cli_integration_4")
```
thinhha, line 87: Could you please change path name clouddq_test_cli_integration_4 to something else?

tests/integration/test_dq_rules.py
```diff
@@ -0,0 +1,153 @@
...
+            temp_dir.mkdir(parents=True)
+            with working_directory(temp_dir):
+                logger.info(f"test_last_modified_in_dq_summary {gcp_application_credentials}")
+                target_table = f"{gcp_project_id}.{target_bq_result_dataset_name}.{target_bq_result_table_name}"
+                args = [
+                    "T1_DQ_1_VALUE_NOT_NULL,T2_DQ_1_EMAIL,T3_DQ_1_EMAIL_DUPLICATE",
```
thinhha, line 93: This should be ALL so we can add more rules later

tests/integration/test_dq_rules.py
```diff
@@ -0,0 +1,153 @@
...
+                    f"{temp_configs_dir}",
+                    f"--gcp_project_id={gcp_project_id}",
+                    f"--gcp_bq_dataset_id={gcp_bq_dataset}",
+                    f"--gcp_region_id={gcp_bq_region}",
+                    f"--target_bigquery_summary_table={target_table}",
+                    "--debug",
+                    "--summary_to_stdout",
```
thinhha, line 100: I think --summary_to_stdout is not needed for the test as the test should already be printing out test failure records no? --summary_to_stdout prints out ALL summary outputs whether it succeeded or failed.

tests/integration/test_dq_rules.py
```diff
@@ -0,0 +1,153 @@
...
+                    dbt_profiles_dir=None,
+                    environment_target="Dev",
+                    gcp_project_id=gcp_project_id,
+                    gcp_region_id=gcp_bq_region,
+                    gcp_bq_dataset_id=gcp_bq_dataset,
+                    gcp_service_account_key_path=gcp_sa_key,
+                    gcp_impersonation_credentials=gcp_impersonation_credentials,
+                )
+                dbt_path = dbt_runner.get_dbt_path()
+                invocation_id = get_dbt_invocation_id(dbt_path)
```
thinhha, line 118: interesting that you had to do this to get the invocation id -- are there any risk that you might retrieve the wrong invocation id here?

tests/integration/test_dq_rules.py
```diff
@@ -0,0 +1,153 @@
...
+                    success_count, failed_count, null_count
+                     FROM `{gcp_project_id}.{target_bq_result_dataset_name}.{target_bq_result_table_name}_expected`;
+                    """
+                    print(f"Dq rules validation query string is \n {sql}")
+                    query_job = client.execute_query(sql)
+                    results = query_job.result()
+                    logger.info("Query done")
+                    rows = list(results)
+                    print(f"Query execution returned {len(rows)} rows")
+                    assert len(rows) == 0
```
thinhha, line 142: does this also print out the error rows if len(rows) is > 0?

tests/integration/test_dq_rules.py
```diff
@@ -0,0 +1,153 @@
...
+
+        try:
+            temp_dir = Path(tmp_path).joinpath("clouddq_test_cli_integration_4")
+            temp_dir.mkdir(parents=True)
+            with working_directory(temp_dir):
+                logger.info(f"test_last_modified_in_dq_summary {gcp_application_credentials}")
+                target_table = f"{gcp_project_id}.{target_bq_result_dataset_name}.{target_bq_result_table_name}"
+                args = [
+                    "T1_DQ_1_VALUE_NOT_NULL,T2_DQ_1_EMAIL,T3_DQ_1_EMAIL_DUPLICATE",
+                    f"{temp_configs_dir}",
```
thinhha, line 94: can we make explicit which config while was used in this test? Ideally, this config file should be created as a part of this test so that if anything fails we can easily figure out which YAML file was used for the test.

Ideally it is a single YAML file being tested rather than a directory & we should print out the YAML file as part of the test so we can inspect it immediately in case of test failures.

thinhha (changes requested):
Thanks for this Amandeep. Looks good overall.

I added a few suggestions to change to make the test more maintainable & a few questions. Can you respond and make the corresponding changes?

tests/integration/test_dq_rules.py
thinhha: we need the test to print out all failing rows in case the assertion fails

tests/integration/test_dq_rules.py
AmandeepSinghCS: right. I did put for viewing summary output. removed it.

tests/integration/test_dq_rules.py
AmandeepSinghCS: I can't think of any risk for now, as it will get the invocation Id as the dbt run completes successfully. We can discuss in details.

tests/integration/test_dq_rules.py
AmandeepSinghCS: > does this also print out the error rows if len(rows) is > 0?

No, for now it doesn't print out the error rows

AmandeepSinghCS: > we need the test to print out all failing rows in case the assertion fails

Will add it in the test.

thinhha (approved):
LGTM - thanks!

PR #116, opened 2021-12-10
better error

thinhha:
This fixes:

1. DataplexClient.get_dataplex_entity() error currently uses the client project-id instead of the project-id provided in the entity_uri
2. add an example rules_dimensions: node to the error message returned by DqRule.validate()
3. EntityUri.from_uri() reports a better error message if entity_uri does not contain the string '://'
4. DqEntity.from_dict() reports a better error message if the input 'metadata' field is not of type dict.
5. DqEntity.resolve_column_config() reports a better error message if the 'column_id' is not found in the entity columns.

### DIFF
--- a/clouddq/classes/dq_configs_cache.py
+++ b/clouddq/classes/dq_configs_cache.py
@@ -185,10 +185,16 @@ class DqConfigsCache:
                     f"Calling Dataplex Metadata list entities API to retrieve schema "
                     f"for entity_uri:\n{pformat(record)}"
                 )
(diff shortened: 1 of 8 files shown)


pbalm:
OK, LGTM.

One thought: If you have to remember to `upper()` the `column_id` and `table_id` in `dq_rule_binding` before using them, maybe we should convert them to uppercase when we initialize the class. But I think we would have to implement a constructor then, to make sure it's always done...

pbalm (approved):

PR #117, opened 2021-12-10
update high watermark logic and return 0 rows

thinhha:
Fix the following bugs:
1. Incremental rule-bindings with 'incremental_time_filter_column_id' will fail if executed for the first time on a BQ dataset and the dq_summary table has not created yet. In the new behaviour, CloudDQ checks if dq_summary table exists and only runs the high-watermark query if the dq_summary table exists, otherwise it executes a full-table scan as it would normally if 'incremental_time_filter_column_id' is not set.
2. Rule-bindings with 'incremental_time_filter_column_id' returns 0 records to dq_summary if no new data has arrived since the last high-watermark. In the new behaviour, CloudDQ returns 1 record with 0 rows_validated and all other statistics set to NULL.
3. CUSTOM_SQL_STATEMENT returns 0 records to dq_summary if the custom-sql returns 0 records (i.e. everything succeeds. In the new behaviour, CloudDQ returns 1 record with 0 rows_validated and all other statistics set to NULL.
4. NOT_NULL rules returning the same count across failed_count and null_count, causing the sum of success_count + failed_count + null_count to exceed the rows_validated.  In the new behaviour, if the rule is NOT_NULL, null_count is set to NULL. 

Going forward:
- success_count + failed_count + null_count should always be equal to rows_validated 
- CloudDQ always return 1 record for each rule-binding/rule into dq_summary
- null_count will be NULL for NOT_NULL rule type
- CUSTOM_SQL_STATEMENT rules will have NULL values in success_count, failed_count, and null_count. And will only populate the existing columns 'complex_rule_validation_errors_count' with the number of rows returned by the custom_sql and additionally a new column 'complex_rule_validation_success_flag' which is TRUE if 'complex_rule_validation_errors_count' is 0, FALSE if 'complex_rule_validation_errors_count' is greater than 0, and NULL if the rule_type if not CUSTOM_SQL_STATEMENT.
- CUSTOM_SQL_STATEMENT is no longer recommended for record-level validation. Users are recommended to use CUSTOM_SQL_EXPR to implement custom record-level this requirements.

This constitutes a breaking change to the dq_summary results & we hope that the change will reduce the confusions in the way different summary statistics are calculated.


hejnal (approved):
LGTM.
Apart from the failing test (it's probably a false alarm) the logic about high watermark and null count handling seems right to me. As well as some minor improvements (error msgs, etc). It would be good to reference in the PR here some of the details about the CUSTOM_SQL_QUERY logic in relation to the null_count logic, to be easier to understand.

clouddq/classes/dq_rule_binding.py
```diff
@@ -197,6 +197,16 @@ def resolve_table_entity_config(
             )
         return table_entity
 
+    def resolve_rule_sql_expr(self: DqRuleBinding, rule: DqRule) -> DqRule:
```
pbalm, line 200: this function doesn't return a DqRule

pbalm (changes requested):
just one thing here

clouddq/classes/dq_rule_binding.py
pbalm: Agree to fix in later PR

pbalm (approved):

PR #121, opened 2021-12-23
Validation error of CUSTOM_SQL_STATEMENT without custom parameter

pbalm:
Currently a rule of type CUSTOM_SQL_STATEMENT cannot be used if it has no custom SQL parameters, because the validation fails. The error is:

```
ValueError: Failed to resolve Rule Binding ID 'UNIQUENESS' with error:
Failed to resolve rule_id 'UNIQUE' in rule_binding_id 'UNIQUENESS' with error:
'NoneType' object is not subscriptable
```

This PR fixes the problem and adds a test.

Note that if `params` is a dict containing a key `rule_binding_arguments` with a value of `None`, then `params.get('rule_binding_arguments', {})` will not return an empty dict, but None.

### DIFF
--- a/clouddq/classes/rule_type.py
+++ b/clouddq/classes/rule_type.py
@@ -120,9 +120,11 @@ def to_sql_custom_sql_statement(params: dict) -> Template:
                     f"custom_sql_arguments '{argument}' not found in "
                     f"input parameters:\n {params}"
                 )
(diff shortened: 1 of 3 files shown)


thinhha (approved):
LGTM - thanks Paul!

PR #124, opened 2022-01-05
allow parametrization of CUSTOM_SQL_EXPR rules

thinhha:


clouddq/classes/rule_type.py
```diff
@@ -84,6 +84,26 @@ def to_sql_custom_sql_expr(params: dict) -> Template:
         f"'custom_sql_expr' in 'params'.\n"
         f"Current value: {custom_sql_expr}",
     )
+    custom_sql_arguments = params.get("custom_sql_arguments", "")
+    if custom_sql_arguments:
+        for argument in custom_sql_arguments:
+            if f"${argument}" not in custom_sql_expr:
+                raise ValueError(
```
andrew-ii, line 91: Should this be error or warning? Not used parameters might indicate something is not as intended but not necessarily be a fatal problem (similar to non-used function parameters in mainstream languages)

tests/resources/expected_configs.json
```diff
@@ -88,6 +94,7 @@
       "NOT_NULL_SIMPLE",
       "REGEX_VALID_EMAIL",
       "CUSTOM_SQL_LENGTH_LE_30",
+      {"CUSTOM_SQL_LENGTH_LE_PARAMETRIZED": {"upper_bound": 0}},
```
andrew-ii, line 97: I might be misunderstanding this, but shouldn't the parameter AST be `{"upper_bound": 30}` to match `tests/resources/dq_rules_configs.yml`?

tests/resources/test_render_run_dq_main_sql_expected.sql
```diff
@@ -125,6 +125,33 @@ SELECT
...
+    'CUSTOM_SQL_LENGTH_LE_PARAMETRIZED' AS rule_id,
+    
+    '<your_gcp_project_id>.<your_bigquery_dataset_id>.contact_details' AS table_id,
+    'value' AS column_id,
+    data.value AS column_value,
+    CAST(NULL AS STRING) AS dimension,
+    CASE
+
+      WHEN value IS NULL THEN CAST(NULL AS BOOLEAN)
+      WHEN LENGTH( value ) <= 0 THEN TRUE
```
andrew-ii, line 141: same comment here, should it not be `WHEN LENGTH( value ) <= 30`?

clouddq/classes/rule_type.py
hejnal: It's an interesting question. I think if no error is thrown (but rather warning) there's risk of leaving behind a hardcoded parameter. It can happen that while developing rules (and testing them directly in the query engine) and then passing them to the yaml file, there's a risk of keeping the hardcoded parameters, so a failure (e.g. when tests) might be a desirable. But I can see the other point of, not being so conservative and letting the execution go.

In any case, the same should apply to CUSTOM_SQL_EXPRESSION.

clouddq/classes/rule_type.py
```diff
@@ -112,13 +132,13 @@ def to_sql_custom_sql_statement(params: dict) -> Template:
                 raise ValueError(
                     f"RuleType: {RuleType.CUSTOM_SQL_STATEMENT} "
                     f"custom_sql_arguments '${argument}' not found in "
-                    f"custom_sql_statement:\n {custom_sql_statement}"
+                    f"custom_sql_statement:\n {pformat(custom_sql_statement)}"
                 )
-            if not params.get("rule_binding_arguments", {}).get(argument, None):
+            if params.get("rule_binding_arguments", {}).get(argument, None) is None:
```
hejnal, line 137: Very helpful bugfix (I came across this issue recently). Shouldn't it have a unit test case, e.g. with 
param_value=0, param_value="", param_value= 
?

tests/resources/expected_configs.json
hejnal: +1

thinhha: I wanted to have a different test for 0 valued arguments. This resource is part of is a unit test. The `dq_rules_configs.yml` file is part of a integration test so there I wanted it to match the output of a non-parametrized rule.

tests/resources/test_render_run_dq_main_sql_expected.sql
thinhha: I wanted to have a different test for 0 valued arguments. This resource is part of is a unit test. The `dq_rules_configs.yml` file is part of a integration test so there I wanted it to match the output of a non-parametrized rule.

clouddq/classes/rule_type.py
thinhha: see above :)

tests/unit/test_classes.py
```diff
@@ -299,6 +299,23 @@ def test_rule_type_custom_to_sql(self):
...
+        """ """
+        params = {
+            "custom_sql_expr": "length(column_name) < $upper_bound and length(column_name) > $lower_bound",
+            "custom_sql_arguments": [
+                "upper_bound",
+                "lower_bound",
+            ],
+            "rule_binding_arguments": {
+                "upper_bound": 10,
+                "lower_bound": 0,
```
thinhha, line 312: @hejnal see the unit test here

clouddq/classes/rule_type.py
thinhha: I would rather be strict here to avoid confusions. We can easily relax restrictions in the future, but we cannot as easily add restrictions that did not exist before.

When we look at adding curried parameters we can relax this.

tests/unit/test_classes.py
hejnal: great, cheers!

clouddq/classes/rule_type.py
hejnal: This works for me.

hejnal:
LGTM, Approved

hejnal (approved):
Thanks for the extra comments, and unit test. LGTM

clouddq/classes/rule_type.py
andrew-ii: agree we can relax later as a non-breaking change so let's go ahead with a warning here 👍

tests/resources/expected_configs.json
andrew-ii: sounds good

PR #126, opened 2022-01-10
Dataproc image version upgrade

hejnal:
Upgrading to a recent minor version of dataproc image not impacted by log4j 2 vulnerability.

More details: https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-release-2.0


hejnal:
> In which case maybe it would be better to leave it as `2.0-ubuntu18` so users will always end up using the latest sub-minor version?

It's great you found that reference, I was suspecting that this might be the case. It's better to leave it than as it is. 

Cheers!

hejnal:
closing the PR, as not implemented (works as intended)

PR #127, opened 2022-01-17
Fixing an issue with conflicting names: column_id=data with the CTE data alias

hejnal:
This issue is specifically related to CUSTOM_SQL_EXPR which gets the column name (with no alias) and then uses it in the sql expression, e.g. REGEX check will only accepts the param of pattern. The substitution of column_id is done by the framework. In other cases, like CUSTOM_SQL_STATEMENT, the 'data' alias can be added to the sql, so there's no code change needed.

### DIFF
--- a/clouddq/classes/dq_rule_binding.py
+++ b/clouddq/classes/dq_rule_binding.py
@@ -304,7 +304,9 @@ class DqRuleBinding:
                     rule_id = rule_id.upper()
                     rule_sql_expr = Template(
                         rule_config["rule_sql_expr"]
-                    ).safe_substitute(column=column_configs.column_name)
+                    ).safe_substitute(column=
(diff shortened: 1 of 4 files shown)


pbalm:
Builds failed due to "Quota exceeded for quota metric 'Read requests for Dataplex Metadata". I hit "rerun", let's see if they pass now.

pbalm:
Now it has a test failure, one of those that compares to the expected JSON output I think: `test_prepare_configs_from_rule_binding`

pbalm (approved):
LGTM

thinhha:
Merging despite test failures because the metadata read quota issues have nothing to do with this change. The Cloud Build tests are more comprehensive than the github action tests so I think we're good to merge.

hejnal:
great, thanks folks!

PR #129, opened 2022-01-20
bq client supports different regions

thinhha:
Fixes a bug with BQ Client validation using the current project id to validate datasets in different regions.


pbalm:
I kind of hate to be that guy, but this change seems a little involved (more than a few lines, and across a good number of different functions) -- would it be possible to create a test for this?

thinhha:
Hi @pbalm, I added a few new tests in `test_bigquery_client` they are passing now.

The metadata tests are still failing however it's unrelated to this change so I'm inclined to ignore them.

Could you please take a look and approve if you can?

pbalm (approved):
Nice! LGTM

thinhha:
awesome thanks!

PR #132, opened 2022-01-25
add throttling for dataplex client

thinhha:


clouddq/integration/dataplex/dataplex_client.py
```diff
@@ -234,31 +249,7 @@ def delete_dataplex_task(
         )
         return response
 
-    def get_dataplex_iam_permissions(
```
hejnal, line 237: Is that clean-up intended?

I can still see some references to that function in the clouddq_dataplex.py

def get_iam_permissions(self) -> list:
        body = {"resource": "dataplex", "permissions": ["roles/dataproc.worker"]}
        return self._client.**get_dataplex_iam_permissions**(
            body=body,
        )

Please doublecheck.

.github/workflows/build-test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
-      max-parallel: 2
+      max-parallel: 1
```
hejnal, line 32: I guess that would impact the test performance overall (not just dataplex related tasks). Are we keeping track of this, so in the future it can be increased again?

scripts/test_pip_install.sh
```diff
@@ -106,7 +107,7 @@ python3 -m clouddq ALL configs \
 
 if [[ -f "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
     # test clouddq with exported service account key
-    python3 -m clouddq ALL configs \
+    python3 -m clouddq T1_DQ_1_VALUE_NOT_NULL,T2_DQ_1_EMAIL,T3_DQ_1_EMAIL_DUPLICATE configs \
```
hejnal, line 110: Is this change related to the throttling issue? If not, why would you need to explicitly set the bindings to be run?

clouddq/integration/dataplex/dataplex_client.py
thinhha: oh right thanks! That's not actually used anywhere so I will delete that as well.

scripts/test_pip_install.sh
thinhha: Yes because ALL includes rule_bindings with entity_uris which will trigger the dataplex metadata API calls, leading to run failures.

hejnal: ok, gotcha! makes sense

.github/workflows/build-test.yml
hejnal: ok, understood!

clouddq/integration/dataplex/dataplex_client.py
hejnal: perfect, thanks!

PR #134, opened 2022-01-25
allow cache to update on new fields

thinhha:


pbalm:
New fields can appear when a batch of rules are ingested into the backing DB, for example that do not have "dimension", the schema created in the DB will not have "dimension". If the next rule has "dimension", we need to be able to update the schema.

pbalm (approved):
LGTM

PR #137, opened 2022-01-31
aggregate entity-level summary tables to avoid bigquery complexity limit

thinhha:
CloudDQ now materializes the dq_summary results as a table at the entity level before unioning the entity-level summaries into the dq_summary table. The entity summary table is created in the same dataset as the rule_binding views and dq_summary table.

The high-level algorithm is:
for each unique (table_id, column_id, row_filter_id), for each batches of roughly 50 rules, create a new entity_summary table aggregating over the rule_bindings.


clouddq/runners/dbt/dbt_connection_configs.py
```diff
@@ -81,7 +81,7 @@ class GcpDbtConnectionConfig(DbtConnectionConfig):
     connection_method: DbtBigQueryConnectionMethod = None
     gcp_service_account_key_path: Optional[str] = None
     service_account_gcp_impersonation_credentials: Optional[str] = None
-    threads: int = 1
+    threads: int = 4
```
hejnal, line 84: Just as a curiosity, I found in the dbt documentation, that the recommended num of threads is 4, as this change goes. But was there a reason to have a single thread only in the previous state?

clouddq/templates/dbt/dbt_project.yml
```diff
@@ -42,9 +42,13 @@ clean-targets: # directories to be removed by `dbt clean`
 models:
   data_quality:
     data_quality_engine:
-      materialized: table
+      +materialized: ephemeral
```
hejnal, line 45: I'm a bit confused by the "ephemeral" type here. This model relates to the dq_summary table which is materialised as a table (not expiring), right? Would the ephemeral translate to the TEMPORAL table in BQ, as opposed to materialised table with expiration?

clouddq/templates/dbt/dbt_project.yml
```diff
@@ -42,9 +42,13 @@ clean-targets: # directories to be removed by `dbt clean`
...
     rule_binding_views:
-      materialized: view
+      +materialized: view
+    entity_dq_statistics:
```
hejnal, line 48: Also here, after doing a few tests I realised that sometimes as a result of the execution (either test or normal run) I see the aggregated statistics object as table (with 6h expiration), but in other cases as a view (e.g. name dev:<project_name>.<dataset>__clouddq_integration_tests_input_data__ingestion_day_level__INVOICE_ID__NONE_1 ). I can see the env in the path name, would that depend on the dev? Or there are other factors?

clouddq/templates/dbt/dbt_project.yml
thinhha: ephemeral is effectively a CTE. This config here defines the default table materialization config, which can be overridden in the model file itself. Effectively, we have two models in data_quality_engine: main.sql which is a CTE/ephemeral materialization, and dq_summary.sql which is an incremental materialization.

clouddq/templates/dbt/dbt_project.yml
thinhha: hmm the env should not appear in the path name. I'll ping you to double check this.

tests/integration/test_dataplex_integration_performance.py
```diff
@@ -0,0 +1,83 @@
...
+        gcp_dataplex_region,
+        gcp_dataplex_lake_name,
+        gcp_dataplex_zone_id,
+        gcp_bq_dataset,
+        target_bq_result_dataset_name,
+        target_bq_result_table_name):
+        try:
+            temp_dir = Path(tmp_path).joinpath("clouddq_test_cli_16_rb_10_rules")
+            temp_dir.mkdir(parents=True)
+            source_configs = Path("tests").joinpath("resources", "configs_16_rb_10_rules.yml").absolute()
```
hejnal, line 51: Just wondering, is there any upper limit of num of different bindings (as combination of entity_column_filter, etc)? In other words, how many different tables/views are allowed to be unioned in the final step?

hejnal (commented):
I ran some tests with those changes applied and it works as expected. I have left a few minor comments, mostly asking about clarifications, reasonale for the change. It is a very important change and thanks for prioritising this Thin.

clouddq/templates/dbt/dbt_project.yml
hejnal: Oh, yes, you're right. My mistake while copying - pasting multiline view_id, there's no ENV in the view name. But I still see a view instead of the table (with expiration), which I would expect..

clouddq/templates/dbt/models/data_quality_engine/main.sql
```diff
@@ -17,7 +17,7 @@
     materialized = 'ephemeral',
```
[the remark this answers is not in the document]
thinhha: technically no

tests/integration/test_dataplex_integration_performance.py
```diff
@@ -0,0 +1,83 @@
...
+            with working_directory(temp_dir):
+                args = [
+                    "ALL",
+                    f"{configs_16_rb_10_rules}",
+                    f"--gcp_project_id={gcp_project_id}",
+                    f"--gcp_bq_dataset_id={gcp_bq_dataset}",
+                    f"--target_bigquery_summary_table={target_table}",
+                    "--enable_experimental_dataplex_gcs_validation",
+                    "--enable_experimental_bigquery_entity_uris",
+                    "--debug",
```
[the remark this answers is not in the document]
thinhha: --debug flag is purely to print additional diagnostic information to help debugging!!
I didn't realize there could be performance problems when running without --debug, could you elaborate?

tests/integration/test_dataplex_integration_performance.py
thinhha: there is an upper limit on the final union step. I think in the orders of 100s of aggregate table, each may contain ~50 rules. So the current upper limit is roughly ~5000 rules.

clouddq/templates/dbt/models/data_quality_engine/main.sql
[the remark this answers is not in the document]
thinhha: It also doesn't hurt to keep this here IMO

clouddq/templates/dbt/dbt_project.yml
thinhha: thanks for confirming! I wonder why this is being created as a view 🤔. technically all models in the `entity_dq_statistics` path should be materialized as a table. Could you share the file path tree of the resulting dbt folder?

tests/integration/test_dataplex_integration_performance.py
hejnal: ok, thanks for the clarification!

clouddq/templates/dbt/models/data_quality_engine/main.sql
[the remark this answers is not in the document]
hejnal: ok, thanks

hejnal (approved):
LGTM! Thanks for putting this together!

clouddq/templates/dbt/dbt_project.yml
hejnal: ok, thanks!

clouddq/templates/dbt/dbt_project.yml
hejnal: Confirmed, it's now creating a table (with expiration), I was using the old dbt config, it worked as expected after cleaning the dbt folder up.

tests/integration/test_dataplex_integration_performance.py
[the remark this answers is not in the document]
hejnal: taken offline, not related to this PR.

PR #148, opened 2022-02-08
clarify that entities configs are not needed if users only use entity_uri

thinhha:


pbalm (changes requested):
If we're adding this (which I support), I think we should also add the dataplex columns to the description of the output in [OVERVIEW.md](OVERVIEW.md#consuming-clouddq-outputs)

thinhha:
Hey @pbalm thanks for your review. I actually didn't realize that we haven't included the dataplex related information there!

I added the relevant columns. PTAL.

pbalm (approved):

PR #153, opened 2022-02-14
Bigquery Native Tables support

AmandeepSinghCS:


clouddq/classes/dq_configs_cache.py
```diff
@@ -461,42 +461,38 @@ def _resolve_bigquery_entity_uri(
...
                     f"'{argument}' is a required argument to look-up metadata for the entity_uri "
-                    "using Dataplex Metadata API.\n"
-                    "Ensure the BigQuery dataset containing this table "
-                    "is registered as an asset in Dataplex.\n"
-                    "You can then specify the corresponding Dataplex "
-                    "projects/locations/lakes/zones as part of the "
-                    "metadata_default_registries YAML configs, e.g.\n"
-                    f"{SAMPLE_DEFAULT_REGISTRIES_YAML}"
+                    "using Bigquery API.\n"
+                    "Ensure the BigQuery dataset contains this table."
```
nauh, line 473: Remove this sentence? This seems a bit unrelated if the error is about missing "project" or "datasets".

clouddq/classes/dq_configs_cache.py
```diff
@@ -461,42 +461,38 @@ def _resolve_bigquery_entity_uri(
...
-                f"{pformat(entity_uri.to_dict())}\n\n"
-            )
-        dataplex_entity = dataplex_entities_match[0]
-        clouddq_entity = dq_entity.DqEntity.from_dataplex_entity(
-            entity_id=entity_uri.get_db_primary_key(),
-            dataplex_entity=dataplex_entity,
-        )
-        return clouddq_entity
+            else:
+                project_id = entity_uri.get_project_id_from_uri()
```
nauh, line 476: Just `entity_uri.configs_dict.get("projects")`, without adding another method in dq_entity_uri.py?

clouddq/integration/bigquery/bigquery_client.py
```diff
@@ -208,3 +208,60 @@ def execute_query(
...
+                        partition_field["type"] = columns[
+                            time_partitioning_field.upper()
+                        ]["type"]
+                        partition_field["partitioning_type"] = time_partitioning_type
+                        partition_fields.append(partition_field)
+
+                if range_partitioning:
+                    raise NotImplementedError(
+                        logger.debug(f"Range Partitioning is not supported currently.")
+                    )
```
nauh, line 251:
```suggestion
                time_partitioning = table_ref.time_partitioning
                if time_partitioning:
                    partition_field = {}
                    time_partitioning_type = time_partitioning.type_
                    time_partitioning_field = time_partitioning.field
                    if not time_partitioning_field:
                        partition_fields = "_PARTITIONTIME"
                    else:
                        partition_field["name"] = time_partitioning_field
                        partition_field["type"] = columns[
                            time_partitioning_field.upper()
                        ]["type"]
                        partition_field["partitioning_type"] = time_partitioning_type
                        partition_fields.append(partition_field)

                range_partitioning = table_ref.range_partitioning
                if range_partitioning:
                    raise NotImplementedError(
                        logger.debug(f"Range Partitioning is not supported currently.")
                    )
```

clouddq/integration/bigquery/bigquery_client.py
```diff
@@ -208,3 +208,60 @@ def execute_query(
...
+                    time_partitioning_field = time_partitioning.field
+                    if not time_partitioning_field:
+                        partition_fields = "_PARTITIONTIME"
+                    else:
+                        partition_field["name"] = time_partitioning_field
+                        partition_field["type"] = columns[
+                            time_partitioning_field.upper()
+                        ]["type"]
+                        partition_field["partitioning_type"] = time_partitioning_type
+                        partition_fields.append(partition_field)
```
nauh, line 246: Can there be multiple partition fields for a table? (wondering why it's a list)

clouddq/integration/bigquery/bigquery_client.py
```diff
@@ -208,3 +208,60 @@ def execute_query(
...
+                    "columns": columns,
+                    "partition_fields": None,
+                }
+
+            logger.debug(f"Schema for table {table} is: {columns_dict}")
+            return columns_dict
+
+        except KeyError as error:
+            logger.fatal(f"Input table `{table}` is not valid.", exc_info=True)
+            raise KeyError(f"\n\nInput table `{table}` is not valid.\n{error}")
```
nauh, line 267: Can this error be thrown from multiple lines in the try block or only from one line? If it's the latter, can we only put that line in a try-except block?

tests/integration/test_dataplex_metadata_uri_templates.py
```diff
@@ -152,14 +152,24 @@ def test_render_run_dq_main_sql(
...
+                output = output.replace(gcp_dataplex_lake_name, "<your_dataplex_lake_id>")
+            else:
+                output = output.replace("CAST(NULL AS STRING) AS dataplex_lake",
+                                        "'<your_dataplex_lake_id>' AS dataplex_lake")
+
+            output = output.replace(rule_binding_id, "<rule_binding_id>")
             output = re.sub(RE_NEWLINES, '\n', output).strip()
             output = re.sub(RE_CONFIGS_HASHSUM, CONFIGS_HASHSUM_REP, output)
             output = re.sub(RE_ASSET_ID, ASSET_ID_REP, output)
+            output = output.replace("CAST(NULL AS STRING) AS dataplex_asset_id,", ASSET_ID_REP)
```
nauh, line 171:
```suggestion
            output = output.replace("CAST(NULL AS STRING) AS dataplex_asset_id,", ASSET_ID_REP)

            output = output.replace(rule_binding_id, "<rule_binding_id>")
            output = re.sub(RE_NEWLINES, '\n', output).strip()
            output = re.sub(RE_CONFIGS_HASHSUM, CONFIGS_HASHSUM_REP, output)
            output = re.sub(RE_ASSET_ID, ASSET_ID_REP, output)
```

nauh (commented):
Thanks Amandeep! I left a few minor comments.

clouddq/integration/bigquery/bigquery_client.py
AmandeepSinghCS: Hive supports multiple partition fields, Metadata API returns a List for Partition Fields . For consistency I have kept the same as we transform them to entity object which can be GCS or Bigquery.

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: done

clouddq/integration/bigquery/bigquery_client.py
AmandeepSinghCS: incorporated the suggestion.

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: done

clouddq/integration/bigquery/bigquery_client.py
AmandeepSinghCS: it will be thrown from one line. incorporated the change.

clouddq/classes/dq_configs_cache.py
```diff
@@ -461,42 +461,37 @@ def _resolve_bigquery_entity_uri(
...
-                f"{pformat(entity_uri.to_dict())}\n\n"
-            )
-        dataplex_entity = dataplex_entities_match[0]
-        clouddq_entity = dq_entity.DqEntity.from_dataplex_entity(
-            entity_id=entity_uri.get_db_primary_key(),
-            dataplex_entity=dataplex_entity,
-        )
-        return clouddq_entity
+            else:
+                project_id = entity_uri.configs_dict.get("projects")
```
thinhha, line 475: s/entity_uri.configs_dict.get("projects")/entity_uri.get_configs("projects")/

clouddq/classes/dq_entity.py
```diff
@@ -362,6 +362,31 @@ def from_dataplex_entity(
                 f"is unsupported for entity:\n {dataplex_entity.to_dict()}"
             )
 
+    @classmethod
+    def from_bq_table_configs(
```
thinhha, line 366: this should take the bigquery entity_uri directly.

clouddq/classes/dq_configs_cache.py
```diff
@@ -461,42 +461,37 @@ def _resolve_bigquery_entity_uri(
...
+                        table=table_name, project_id=project_id
+                    )
+                    clouddq_entity = dq_entity.DqEntity.from_bq_table_configs(
+                        entity_id=entity_uri.get_entity_id(),
+                        bq_configs=configs,
+                        columns_dict=columns_dict,
+                    )
+                    return clouddq_entity
+                else:
+                    raise RuntimeError(f"Bigquery Table {table_name} does not exists")
```
thinhha, line 497: please add entity_uri into error log

clouddq/integration/bigquery/bigquery_client.py
```diff
@@ -208,3 +208,59 @@ def execute_query(
...
+                    partition_field["type"] = columns[time_partitioning_field.upper()][
+                        "type"
+                    ]
+                    partition_field["partitioning_type"] = time_partitioning_type
+                    partition_fields.append(partition_field)
+
+            range_partitioning = table_ref.range_partitioning
+            if range_partitioning:
+                raise NotImplementedError(
+                    logger.debug(f"Range Partitioning is not supported currently.")
```
thinhha, line 253: please add more descriptive error message. e.g. BigQuery range partition for table <> is not supported.

thinhha (changes requested):

clouddq/integration/bigquery/bigquery_client.py
AmandeepSinghCS: done

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: done

clouddq/classes/dq_configs_cache.py
```diff
@@ -461,42 +461,37 @@ def _resolve_bigquery_entity_uri(
...
+            else:
+                project_id = entity_uri.configs_dict.get("projects")
+                table_name = entity_uri.get_table_name_from_entity_uri()
+                bq_native_table_exists = bigquery_client.is_table_exists(
+                    table=table_name, project_id=project_id
+                )
+                if bq_native_table_exists:
+                    logger.debug(
+                        f"The Table {table_name} in the "
+                        f"specified entity_uri {entity_uri} \n"
```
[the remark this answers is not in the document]
AmandeepSinghCS: done

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: done

clouddq/classes/dq_entity.py
AmandeepSinghCS: updated

clouddq/classes/dq_configs_cache.py
```diff
@@ -461,42 +461,37 @@ def _resolve_bigquery_entity_uri(
...
+                        f"exists is Bigquery."
+                    )
+                    configs = entity_uri.configs_dict
+                    columns_dict = bigquery_client.get_table_schema(
+                        table=table_name, project_id=project_id
+                    )
+                    clouddq_entity = dq_entity.DqEntity.from_bq_table_configs(
```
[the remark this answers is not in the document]
AmandeepSinghCS: updated

clouddq/classes/dq_configs_cache.py
```diff
@@ -443,7 +443,7 @@ def _resolve_dataplex_entity_uri(
     def _resolve_bigquery_entity_uri(
         self,
         entity_uri: dq_entity_uri.EntityUri,
-        client: clouddq_dataplex.CloudDqDataplexClient,
```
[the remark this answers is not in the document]
AmandeepSinghCS: correct me if I am wrong or if I am missing something here but why do we need to call metadata API, if we are using BQ tables API to look for any table in the dataset that is assigned as a asset to dataplex or not.

clouddq/classes/dq_configs_cache.py
```diff
@@ -440,10 +442,62 @@ def _resolve_dataplex_entity_uri(
...
+        for argument in required_arguments:
+            uri_argument = entity_uri.get_configs(argument)
+            if not uri_argument:
+                logger.info(
+                    f"Failed to retrieve default Dataplex '{argument}' for "
+                    f"entity_uri: {entity_uri.complete_uri_string}. \n"
+                    f"'{argument}' is a required argument to look-up metadata for the entity_uri "
+                    "using Dataplex Metadata API.\n"
+                    "Ensure the BigQuery dataset containing this table "
+                    "is registered as an asset in Dataplex.\n"
```
nauh, line 460: "registered as" -> "attached to"

clouddq/classes/dq_configs_cache.py
```diff
@@ -461,42 +515,42 @@ def _resolve_bigquery_entity_uri(
...
+                        clouddq_entity = dq_entity.DqEntity.from_bq_entity_uri(
+                            entity_id=entity_uri.get_entity_id(),
+                            entity_uri=entity_uri,
+                            bigquery_client=bigquery_client,
+                        )
+                    return clouddq_entity
+                else:
+                    raise RuntimeError(
+                        f"Bigquery Table '{table_name}' specified in the "
+                        f"entity uri '{entity_uri}' does not exists"
```
nauh, line 555: s/exists/exist

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: updated this to=> Ensure the BigQuery dataset containing this table is attached as an asset in Dataplex

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: updated

clouddq/classes/dq_configs_cache.py
```diff
@@ -461,42 +515,42 @@ def _resolve_bigquery_entity_uri(
...
-                f"Parsed entity_uri configs:\n"
-                f"{pformat(entity_uri.to_dict())}\n\n"
-            )
-        dataplex_entity = dataplex_entities_match[0]
-        clouddq_entity = dq_entity.DqEntity.from_dataplex_entity(
-            entity_id=entity_uri.get_db_primary_key(),
-            dataplex_entity=dataplex_entity,
-        )
-        return clouddq_entity
+            else:
```
thinhha, line 528: this else clause is being run inside of the loop `for argument in required_arguments:`, which means that it is executed at least 3 times, and may fail if a uri has `projects` but not `dataset`.

Please take everything after the else clause out of this loop.

clouddq/classes/dq_entity_uri.py
```diff
@@ -179,3 +179,10 @@ def _validate_bigquery_uri_fields(self, config_id: str, configs: dict) -> None:
                     f"input config under headings 'metadata_default_registries', e.g.\n"
                     f"{SAMPLE_DEFAULT_REGISTRIES_YAML}",
                 )
+
+    def get_table_name_from_entity_uri(self: EntityUri):
```
thinhha, line 183: maybe just `get_table_name` is sufficient? the entity_uri part if self-explanatory from the parent class.

clouddq/classes/dq_configs_cache.py
```diff
@@ -440,10 +442,62 @@ def _resolve_dataplex_entity_uri(
...
+                        f"{pformat(entity_uri.to_dict())}\n\n"
+                    )
+                    return False
+                else:
+                    dataplex_entity = dataplex_entities_match[0]
+                    clouddq_entity = dq_entity.DqEntity.from_dataplex_entity(
+                        entity_id=entity_uri.get_db_primary_key(),
+                        dataplex_entity=dataplex_entity,
+                    )
+                    return clouddq_entity
```
thinhha, line 494: I think we have the same bug here as discussed earlier Amandeep. Can you please move this out of the for loop?

clouddq/classes/dq_entity_uri.py
thinhha: have you changed this method name as well @AmandeepSinghCS ?

AmandeepSinghCS: @thinhha yes this is renamed

clouddq/classes/dq_configs_cache.py
```diff
@@ -486,17 +473,81 @@ def _resolve_bigquery_entity_uri(
...
+            clouddq_entity = dq_entity.DqEntity.from_dataplex_entity(
+                entity_id=entity_uri.get_db_primary_key(),
+                dataplex_entity=dataplex_entity,
+            )
         return clouddq_entity
+
+    def _resolve_bigquery_entity_uri(
+        self,
+        entity_uri: dq_entity_uri.EntityUri,
+        client: clouddq_dataplex.CloudDqDataplexClient,
```
nauh, line 496: Can we rename `client` to `dataplex_client`?

clouddq/classes/dq_configs_cache.py
```diff
@@ -486,17 +473,81 @@ def _resolve_bigquery_entity_uri(
...
+        project_id = entity_uri.get_configs("projects")
+        table_name = entity_uri.get_table_name()
+        bq_native_table_exists = bigquery_client.is_table_exists(
+            table=table_name, project_id=project_id
+        )
+        if bq_native_table_exists:
+            logger.debug(
+                f"The Table '{table_name}' in the "
+                f"specified entity_uri '{entity_uri}' "
+                f"exists is Bigquery."
```
nauh, line 535: "exists in BigQuery"?

clouddq/classes/dq_configs_cache.py
```diff
@@ -486,17 +473,81 @@ def _resolve_bigquery_entity_uri(
...
+                raise RuntimeError(
+                    f"Failed to retrieve default Bigquery '{argument}' for "
+                    f"entity_uri: {entity_uri.complete_uri_string}. \n"
+                    f"'{argument}' is a required argument to look-up metadata for the entity_uri "
+                    "using Bigquery API.\n"
+                )
+
+        project_id = entity_uri.get_configs("projects")
+        table_name = entity_uri.get_table_name()
+        bq_native_table_exists = bigquery_client.is_table_exists(
```
nauh, line 528: Is this check for the BigQuery native tables only, or BigQuery native tables and external tables? If the latter, then maybe just call it `bq_table_exists`?

On a related note, it's probably good to clarify what "Bigquery Native Tables Support" entails exactly (regarding external table support).

clouddq/classes/dq_configs_cache.py
```diff
@@ -486,17 +473,81 @@ def _resolve_bigquery_entity_uri(
...
+                f"exists is Bigquery."
+            )
+            dataplex_entity = self.is_dataplex_entity(
+                entity_uri=entity_uri, client=client
+            )
+            if dataplex_entity:
+                clouddq_entity = dataplex_entity
+            else:
+                clouddq_entity = dq_entity.DqEntity.from_bq_entity_uri(
+                    entity_id=entity_uri.get_entity_id(),
```
nauh, line 544: If `entity_id` can be retrieved from `entity_uri`, then we don't need to pass both `entity_id` and `entity_uri` to the `from_bq_entity_uri` method -- only `entity_uri` should suffice.

clouddq/integration/bigquery/bigquery_client.py
```diff
@@ -208,3 +208,61 @@ def execute_query(
...
+                    partition_field["type"] = columns[time_partitioning_field.upper()][
+                        "type"
+                    ]
+                    partition_field["partitioning_type"] = time_partitioning_type
+                    partition_fields.append(partition_field)
+
+            range_partitioning = table_ref.range_partitioning
+            if range_partitioning:
+                raise NotImplementedError(
+                    logger.debug(
```
nauh, line 252: Should this be debug level logging or higher?

clouddq/integration/bigquery/bigquery_client.py
```diff
@@ -208,3 +208,61 @@ def execute_query(
...
+        logger.debug(f"Table Partitioning Type is : {table_partitioning_type}")
+        if table_partitioning_type:
+            partition_fields = []
+            time_partitioning = table_ref.time_partitioning
+            if time_partitioning:
+                partition_field = {}
+                time_partitioning_type = time_partitioning.type_
+                time_partitioning_field = time_partitioning.field
+                if not time_partitioning_field:
+                    partition_fields = "_PARTITIONTIME"
```
nauh, line 240: How is `partition_fields` used? I wonder how we distinguish it being a string (i.e. "_PARTITIONTIME") from it being an array.

clouddq/BUILD
```diff
@@ -56,6 +56,7 @@ py_binary(
         requirement("click"),
         requirement("coloredlogs"),
         requirement("sqlite-utils"),
+        requirement("markupsafe"),
```
nauh, line 59: Just curious, where do we actually use `markupsafe` lib from the code?

tests/resources/configs/rule_bindings/team-7-rule-bindings.yml
```diff
@@ -0,0 +1,32 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+rule_bindings:
+  T12_URI_NATIVE_BQ_NOT_NULL:
+    entity_uri: bigquery://locations/<my-gcp-dataplex-region-id>/lakes/<my-gcp-dataplex-lake-id>/zones/<my-gcp-dataplex-zone-id>/projects/<my-gcp-project-id>/datasets/austin_311/tables/311_service_requests
```
nauh, line 17: I wonder if we should use a public dataset or something created and owned by us.

- Theoretically, public dataset can change (? for example, adding a new column), which can then make our test fail. I don't think we should create such dependency for our tests.
- Public dataset does not provide full coverage of our test scenarios. For example, `partition_fields` is either `None` or `null` for all the testing cases in this PR. I think we should have at least a test case where it's not `None` or `null`.

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: done

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: done

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: It can be native or external table created by dataplex. We can rename it to the bq_table_exists.

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: updated the method definition

clouddq/integration/bigquery/bigquery_client.py
AmandeepSinghCS: This should be only message for NotImplementedError, logger is not needed. Thanks for catching this.

clouddq/BUILD
AmandeepSinghCS: we use jinja, jinja internally has a dependency on markupsafe.

clouddq/integration/bigquery/bigquery_client.py
AmandeepSinghCS: _PARTITIONTIME is  a pseudocolumn for ingestion-time partitioning. The value of this column is the ingestion time for each row, truncated to the partition boundary (such as hourly or daily). This will be the default partition field. In case of metadata for GCS entities, it may have multiple partition fields so a list is used. For custom hive partitioning.

tests/resources/configs/rule_bindings/team-7-rule-bindings.yml
AmandeepSinghCS: 1. Theoretically, public dataset can change (? for example, a new column added), which can then make our test fail. I don't think we should create such dependency for our tests.

> I have made a copy of public dataset in our project, so the table structure is not going to change until we make a change. So our tests will run independent of public dataset.

2.  Public dataset does not provide full coverage of our test scenarios. For example, partition_fields is either None or null for all the testing cases in this PR. I think we should have at least a test case where it's not None or null.

>  Yes, I agree I will add a test for the same.

nauh (approved):
LGTM. Please wait for approval from thinhha as well before merging the changes. Thanks.

clouddq/classes/dq_configs_cache.py
```diff
@@ -486,17 +473,80 @@ def _resolve_bigquery_entity_uri(
...
+            dataplex_entity = self.is_dataplex_entity(
+                entity_uri=entity_uri, client=client
+            )
+            if dataplex_entity:
+                clouddq_entity = dataplex_entity
+            else:
+                clouddq_entity = dq_entity.DqEntity.from_bq_entity_uri(
+                    entity_uri=entity_uri,
+                    bigquery_client=bigquery_client,
+                )
```
thinhha, line 546: Can we please add an integration test for this behaviour?

clouddq/classes/dq_entity.py
```diff
@@ -362,6 +364,38 @@ def from_dataplex_entity(
...
+            "columns": columns_dict.get("columns"),
+            "environment_override": {},
+            "entity_id": entity_id,
+            "dataplex_name": configs.get("dataplex_name", None),
+            "dataplex_lake": configs.get("dataplex_lake", None),
+            "dataplex_zone": configs.get("dataplex_zone", None),
+            "dataplex_location": configs.get("dataplex_location", None),
+            "dataplex_asset_id": configs.get("dataplex_asset_id", None),
+            "dataplex_createTime": configs.get("dataplex_createTime", None),
+            "dataplex_updateTime": configs.get("dataplex_updateTime", None),
```
thinhha, line 393: If the code got to this point, it is safe to say there are no corresponding Dataplex entity_id for the entity_uri, therefore all "dataplex_*" fields should be set to 'None' to ensure correctness.

clouddq/integration/bigquery/bigquery_client.py
```diff
@@ -208,3 +208,59 @@ def execute_query(
         )
 
         return query_job
+
+    def get_table_schema(self, table: str, project_id: str = None) -> dict:
```
thinhha, line 212: Can we please add an integration test for this behaviour?

clouddq/integration/bigquery/bigquery_client.py
```diff
@@ -208,3 +208,59 @@ def execute_query(
...
+                partition_field = {}
+                time_partitioning_type = time_partitioning.type_
+                time_partitioning_field = time_partitioning.field
+                if not time_partitioning_field:
+                    partition_fields = "_PARTITIONTIME"
+                else:
+                    partition_field["name"] = time_partitioning_field
+                    partition_field["type"] = columns[time_partitioning_field.upper()][
+                        "type"
+                    ]
```
thinhha, line 245: Is this effectively "STRING" for all Hive-partitioned tables?

clouddq/integration/bigquery/bigquery_client.py
thinhha: It is intuitive for a field to be either a STRING or an ARRAY. I would prefer if we keep it consistent as an ARRAY in the CloudDQ implementation.

Is there any reason why we cannot force '_PARTITIONTIME' into the ARRAY format? e.g.

    "partition_fields": [{"name": "_PARTITIONTIME",
                          "partitioning_type": "DAY", <- ensure you get the correct partition type of day, hourly, or monthly
                          "type": "TIMESTAMP"}]

thinhha: Could you please also add an integration test for _PARTITIONTIME time-partitioned table?

thinhha (commented):
Adding a few more nits to be addressed in #157

clouddq/integration/bigquery/bigquery_client.py
thinhha: Does BigQuery tell us what the partition duration setting is? e.g. DAY, HOUR, MONTH

AmandeepSinghCS: Yes

clouddq/integration/bigquery/bigquery_client.py
AmandeepSinghCS: added the tests for this behaviour

clouddq/classes/dq_configs_cache.py
AmandeepSinghCS: added the test for this behaviour

clouddq/classes/dq_entity.py
AmandeepSinghCS: Yes I agree. All this should be None by default if the code got to this point.

PR #158, opened 2022-03-17
added a flag for dbt intermediate table expiration hours

AmandeepSinghCS:


clouddq/main.py
```diff
@@ -211,6 +211,13 @@
     is_flag=True,
     default=False,
 )
+@click.option(
+    "--intermediate_table_expiration_hours",
+    help="Configure the DBT created intermediate tables expiration hours"
```
thinhha, line 216: please remove reference to DBT

clouddq/utils.py
```diff
@@ -101,6 +101,26 @@ def write_templated_file_to_path(path: Path, lookup_table: typing.Dict) -> None:
     path.write_text(get_template_file(lookup_table.get(path.name)))
 
 
+def set_dbt_intermediate_expiration_hours(
```
thinhha, line 104: I don't really like this method here. It feels very hacky to replace a string on a file format that was created from a class that contains the field we are trying to replace. Can we not find another way to override the expiration hours on the __init__ method of DbtRunner?

thinhha (changes requested):
Thanks Amandeep. Is there any way we could refactor away set_dbt_intermediate_expiration_hours? I don't think this method should exist.

clouddq/main.py
AmandeepSinghCS: remove DBT reference

clouddq/utils.py
AmandeepSinghCS: We are writing the templated file that contains expiration hours. So I am updating the templated file content before writing to the target location. This processing is done inside **init** method of DbtRunner. We can chat more.

clouddq/runners/dbt/dbt_runner.py
```diff
@@ -213,15 +218,33 @@ def _resolve_dbt_path(
...
                 f"Writing templated file to: {dbt_project_path}/dbt_project.yml"
             )
-            write_templated_file_to_path(dbt_project_path, DBT_TEMPLATED_FILE_LOCATIONS)
-        logger.debug(f"Using 'dbt_project_path': {dbt_project_path}")
+            if intermediate_table_expiration_hours:
+                set_dbt_intermediate_expiration_hours(
+                    path=dbt_project_path,
+                    intermediate_table_expiration_hours=intermediate_table_expiration_hours,
+                    lookup_table=DBT_TEMPLATED_FILE_LOCATIONS
+                )
```
nauh, line 235: It seems like this code snippet will run regardless of whether `dbt_project_path` is a file or not.

Can we refactor this code a little bit?

```
if intermediate_table_expiration_hours:
... ...
else:
    if dbt_project_path.is_file():
        logger.debug(f"Using 'dbt_project_path': {dbt_project_path}")
    else:
        logger.debug(...)
        write_templated_file_to_path(dbt_project_path, DBT_TEMPLATED_FILE_LOCATIONS)
```

clouddq/main.py
```diff
@@ -211,6 +211,14 @@
     is_flag=True,
     default=False,
 )
+@click.option(
+    "--intermediate_table_expiration_hours",
+    help="Configure the bigquery job intermediate tables expiration hours "
+    "in the provided dataset '--gcp_bq_dataset_id' . Expiration time "
+    "defaults to 24 hours if not set. ",
```
nauh, line 218: Remove "Expiration time defaults to 24 hours if not set." and set `show_default` to `true` in `click.option`?

AmandeepSinghCS: incorporated

clouddq/runners/dbt/dbt_runner.py
AmandeepSinghCS: refractored

clouddq/main.py
```diff
@@ -226,6 +234,7 @@ def main(  # noqa: C901
     dry_run: bool,
     progress_watermark: bool,
     target_bigquery_summary_table: Optional[str],
+    intermediate_table_expiration_hours: Optional[int],
```
thinhha, line 237: Please remove Optional

clouddq/runners/dbt/dbt_runner.py
```diff
@@ -213,15 +220,42 @@ def _resolve_dbt_path(
             logger.debug(f"Using 'dbt_path': {dbt_path}")
         return dbt_path
 
-    def _prepare_dbt_project_path(self) -> None:
+    def _prepare_dbt_project_path(self,
+                                  intermediate_table_expiration_hours: Optional[int]) -> None:
```
thinhha, line 224: this is 'int' not 'Optional[int]'

clouddq/runners/dbt/dbt_runner.py
```diff
@@ -213,15 +220,42 @@ def _resolve_dbt_path(
...
+
         dbt_project_path = self.dbt_path.absolute().joinpath("dbt_project.yml")
-        if not dbt_project_path.is_file():
+        if intermediate_table_expiration_hours:
```
thinhha, line 227: if 'intermediate_table_expiration_hours' is not optional then we don't need this.

thinhha: please remove this if clause and reduce indent for everything after it

thinhha (commented):
please remove optional and if clause in _prepare_dbt_project_path

PR #165, opened 2022-04-12
add RECORD type

thinhha:
Fix error `ValueError: 'RECORD' is not a valid DatabaseColumnType` when using entity_uri but not when using entity_id.


AmandeepSinghCS (approved):
LGTM - Thanks!

PR #169, opened 2022-04-26
Target table req arg

AmandeepSinghCS:


clouddq/main.py
```diff
@@ -582,18 +582,17 @@ def main(  # noqa: C901
...
-                    "It is needed to append the dq summary results to the "
-                    "provided target bigquery table. This will become a "
-                    "required argument in v1.0.0"
-                )
                 if summary_to_stdout:
                     logger.warning(
                         "'--summary_to_stdout' was set but does"
                         " not take effect unless "
                         "'--target_bigquery_summary_table' is provided"
                     )
```
nauh, line 590: Does this still make sense? Now --target_bigquery_summary_table is a required parameter anyway.

nauh, line 590: Does this still make sense? Now --target_bigquery_summary_table is a required parameter anyway.

AmandeepSinghCS: yes i agree , I haven't removed it because if we want to give log message, but that doesn't add much value as we are throwing error just after. Removed it now.

tests/integration/test_cli_integration.py
```diff
@@ -395,6 +395,42 @@ def test_cli_dry_run_implicit_region_id(
...
+                    f"--gcp_region_id={gcp_bq_region}",
+                    f"--target_bigquery_summary_table={target_table}",
+                    "--debug",
+                    "--summary_to_stdout",
+                ]
+                logger.info(f"Args: {' '.join(args)}")
+
+                result = runner.invoke(main, args)
+                print(result.output)
+                assert result.exit_code == 1
```
nauh, line 430: Is it possible to assert the error message or something more specific and descriptive?

AmandeepSinghCS: We can assert the instance of error as well. I have added that assertion as well.

clouddq/main.py
```diff
@@ -553,41 +553,42 @@ def main(  # noqa: C901
...
-                                "partition_date": partition_date,
-                                "num_rows_loaded_to_target_table": num_rows,
-                            }
-                        },
-                        cls=JsonEncoderDatetime,
+                if target_bigquery_summary_table == dq_summary_table_name:
+                    raise ValueError(
+                        f"The target bigquery summary table name `{target_bigquery_summary_table}` "
+                        f"must not be same as dq summary table name `{dq_summary_table_name}`"
                     )
```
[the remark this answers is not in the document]
AmandeepSinghCS: updated

clouddq/main.py
```diff
@@ -556,7 +556,9 @@ def main(  # noqa: C901
                 if target_bigquery_summary_table == dq_summary_table_name:
                     raise ValueError(
                         f"The target bigquery summary table name `{target_bigquery_summary_table}` "
-                        f"must not be same as dq summary table name `{dq_summary_table_name}`"
+                        f"cannot be same as dq summary table name `{dq_summary_table_name}` which "
+                        f"is reserved for storing the intermediate results which will be used by "
```
nauh, line 560: nit:

s/storing the intermediate results which will be used by/storing the intermediate results used by

tests/integration/test_cli_integration.py
nauh: Is it possible to check the error message (e.g. "The target bigquery summary table name cannot be the same ..."). I wonder if other issues can also trigger SystemExit

nauh (approved):

clouddq/main.py
AmandeepSinghCS: incorporated

tests/integration/test_cli_integration.py
thinhha: @AmandeepSinghCS you can do
```
                error_message = (
                    "..."
                )
                assert error_message in result.output
```

AmandeepSinghCS: added the assert for error message

nauh (approved):
Thanks @AmandeepSinghCS!

thinhha (approved):
LGTM - thanks @AmandeepSinghCS !

PR #173, opened 2022-05-13
added get high watermark fields and current execution timestamp to co…

AmandeepSinghCS:
added get high watermark fields

### DIFF
--- a/clouddq/lib.py
+++ b/clouddq/lib.py
@@ -15,7 +15,6 @@
 """todo: add lib docstring."""
 from __future__ import annotations
 
-from datetime import datetime
 from pathlib import Path
 from pprint import pformat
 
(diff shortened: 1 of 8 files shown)


clouddq/lib.py
```diff
@@ -195,6 +202,26 @@ def prepare_configs_from_rule_binding_id(
         {"configs_hashsum": sha256_digest(json.dumps(resolved_rule_binding_configs))}
     )
     configs.update({"progress_watermark": progress_watermark})
+    print("Incremental time fileter column")
```
nauh, line 205: Remove these "print" lines? (I assume it's for debugging)

nauh (commented):
Do we have a test case for this? (in addition to the test which checks the expected template)

clouddq/lib.py
AmandeepSinghCS: yes, modified it to logger.

nauh:
> checks

Issue #171, opened 2022-05-04
Allow multiple row_filters and params to row_filters


bogo96:
Is there any reason that not allow multiple row_filters and params?
While using clouddq, i got some needs of this in some situation.

if allow params, then row_filters can be used in general. I mean one row_filter can be used in many entities.
Or sometime, we want to filter by a column other than the one used for rule_binding.
I suggest modifying like below, then filtering email row_filter can be use any entity with any given column.
  
```yaml
row_filters:
  NONE:
    filter_sql_expr: |-
      True

  DATA_TYPE_EMAIL:
    params: |-
      - column
    filter_sql_expr: |-
      $column = 'email'

rule_bindings:
  T2_DQ_1_EMAIL:
    entity_id: TEST_TABLE
    column_id: VALUE
    row_filter_ids:
      - DATA_TYPE_EMAIL:
           column: contact_type
    rule_ids:
      - NOT_NULL_SIMPLE
      - REGEX_VALID_EMAIL
      - CUSTOM_SQL_LENGTH:
          upper_bound: 30
      - NOT_BLANK
    metadata:
      brand: one
```

Is there any way to use it in this situation? if not, can i adding this?

thinhha:
Hey @bogo96 thanks for opening this issue. I agree that we should allow 1) allowing multiple row_filters in a rule_binding and 2) allow parametrization of row_filters.

This would be a fairly involved PR but happy to support you in this if you are up to it!

You need up update the [DqRowFilter](https://github.com/GoogleCloudPlatform/cloud-data-quality/blob/main/clouddq/classes/dq_row_filter.py#L31) class to support taking parameters and adding a method to allow resolving these parameters into a valid SQL string.  See below for example code where we've added parametrization to a CUSTOM_SQL_EXPR rule_type:
https://github.com/GoogleCloudPlatform/cloud-data-quality/blob/main/clouddq/classes/dq_rule_binding.py#L246
https://github.com/GoogleCloudPlatform/cloud-data-quality/pull/124/files#diff-53a624ae4d4f1930d8966d1ddd72980a9e347567cd8b9f05ca3018eb739ebae8

yankisimo:
Hi @thinhha @bogo96, is this issue in progress? blocked? I definitely think that a rule_filter with params could be highly more efficient than the current definition where we need to define a distinct rule_filter for each column_name that we want to use in all tables or redefine entities.

thinhha:
[...]

Issue #70, opened 2021-10-29
Known Issue: Github Actions fails when a PR comes from a fork


thinhha:
Example PR: https://github.com/GoogleCloudPlatform/cloud-data-quality/pull/69

Github's response: https://github.community/t/make-secrets-available-to-builds-of-forks/16166/14

We need to update the CI process to allow testing from forks, otherwise contributors will all need maintainer permission to the project.

PR #174, opened 2022-05-16
Add params to row-filter and modify to be used multiple at the same time

bogo96:
- add params
- allow multi-use
- modify config as changes

### Related Issue
#171


clouddq/classes/dq_rule_binding.py
```diff
@@ -92,13 +92,14 @@ def from_dict(
         )
         if column_id:
             column_id.upper()
-        row_filter_id: str = get_from_dict_and_assert(
+        row_filter_ids: list[str] = get_from_dict_and_assert(
             config_id=rule_binding_id,
             kwargs=kwargs,
-            key="row_filter_id",
+            key="row_filter_ids",
```
thinhha, line 98: I suggest either 1) leaving this field as 'row_filter_id' for backwards compatibility or 2) allowing one-of 'row_filter_id' or 'row_filter_ids'. Perhaps (1) is preferable despite the unfortunate naming.

clouddq/classes/dq_rule_binding.py
```diff
@@ -92,13 +92,14 @@ def from_dict(
...
+            assertion=lambda x: type(x) == list,
+            error_msg=f"Rule Binding ID: '{rule_binding_id}' must have defined value "
+            f"'row_filter_ids' of type 'list'.",
         )
-        if row_filter_id:
-            row_filter_id.upper()
```
thinhha, line 101: I think we'll need to caps all the row_filter_ids as well

clouddq/classes/dq_rule_binding.py
```diff
@@ -251,12 +252,35 @@ def resolve_rule_config_list(
...
-        row_filter = configs_cache.get_row_filter_id(self.row_filter_id.upper())
-        return row_filter
+    ) -> list[DqRowFilter]:
+        resolved_row_filter_config_list = []
+        for row_filter in self.row_filter_ids:
+            if type(row_filter) == dict:
+                if len(row_filter) > 1:
+                    raise ValueError(
+                        f"Rule Binding {self.rule_binding_id} has "
+                        f"invalid configs in rule_ids. "
```
thinhha, line 265: replace "rule_ids" with "row_filter_id"?

clouddq/classes/dq_rule_binding.py
```diff
@@ -251,12 +252,35 @@ def resolve_rule_config_list(
...
+                        f"Each nested row_filter_id objects cannot "
+                        f"have more than one rule_id. "
```
thinhha, line 267: replace "rule_ids" with "row_filter_id"?

thinhha:
Hi @bogo96 thanks for this amazing first contributions!

I've approved for both Github Actions and Cloud Run to run on your PR. Looks like Github Actions is failing because of https://github.com/GoogleCloudPlatform/cloud-data-quality/issues/70 (there is no workaround).

The Cloud Build logs are not directly visible in Github so I need to share them with you here. The tests are currently failing due to:

```
Failed to parse Rule Binding with error:
Config ID: T3_DQ_1_EMAIL_DUPLICATE must define non-empty value: 'row_filter_ids'.
```

This is because of https://github.com/GoogleCloudPlatform/cloud-data-quality/pull/174#discussion_r874154280. What do you think about (2) allow one-of 'row_filter_id' which is a string or 'row_filter_ids' which is a list instead?

BTW wanted to check if you are able to run `make test` from where you are creating this PR?

Thanks again!
Thinh

bogo96:
> Hi @bogo96 thanks for this amazing first contributions!
> 
> I've approved for both Github Actions and Cloud Run to run on your PR. Looks like Github Actions is failing because of #70 (there is no workaround).
> 
> The Cloud Build logs are not directly visible in Github so I need to share them with you here. The tests are currently failing due to:
> 
> ```
> Failed to parse Rule Binding with error:
> Config ID: T3_DQ_1_EMAIL_DUPLICATE must define non-empty value: 'row_filter_ids'.
> ```
> 
> This is because of [#174 (comment)](https://github.com/GoogleCloudPlatform/cloud-data-quality/pull/174#discussion_r874154280). What do you think about (2) allow one-of 'row_filter_id' which is a string or 'row_filter_ids' which is a list instead?
> 
> BTW wanted to check if you are able to run `make test` from where you are creating this PR?
> 
> Thanks again! Thinh

Hi, I really appreciated with your review. I wonder how to `make test` in my local.
I think i have to do something to my gcp project and dataset. Cause once I ran `make test`, there are some fails about `not exist of table which in bigquery-public-data project` even though I made `contact_details` table in my dataset.
Is there any guide about it?

clouddq/classes/dq_rule_binding.py
bogo96: it was in `resolve_row_filter_configs` function like rule-id but now, you can see it L.102-109

clouddq/classes/dq_rule_binding.py
bogo96: i agree:) i fix it to allow both!

thinhha (approved):
LGTM - thanks! I will create a separate PR to create a few more test for this PR before merging to main.

PR #176, opened 2022-06-08
Update composer docs

AmandeepSinghCS:


scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
```diff
@@ -0,0 +1,226 @@
...
+from requests_oauth2 import OAuth2BearerToken
+import json
+import requests
+import time
+
+DATAPLEX_PROJECT_ID = "your-dataplex-project-id"  # The Google Cloud Project where the Dataplex task will be created
+DATAPLEX_REGION = "your-dataplex-region-id"  # The region of the Dataplex Lake where the data quality task will be created.
+DATAPLEX_LAKE_ID = "your-dataplex-lake-id"  # dataplex lake id
+SERVICE_ACC = "service-account-to-execute-task"  # The service account used for executing the task. Ensure this service account has sufficient IAM permissions on your project including BigQuery Data Editor, BigQuery Job User, Dataplex Editor, Dataproc Worker, Service Usage Consumer
+PUBLIC_CLOUDDQ_EXECUTABLE_BUCKET_NAME = "your-public-bucket-with-clouddq-executable-and-hashsum" #Public Cloud Storage bucket containing the prebuilt data quality executable artifact and hashsum. There is one bucket per GCP region.
```
nauh, line 39: "#Public": missing space between "#" and "Public"

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+    # getting the credentials and project details for gcp project
+    credentials, your_project_id = google.auth.default(scopes=["https://www.googleapis.com/auth/cloud-platform"])
+
+    # getting request object
+    auth_req = google.auth.transport.requests.Request()
+
+    print(credentials.valid)  # logger.debugs False
+    credentials.refresh(auth_req)  # refresh token
+    # cehck for valid credentials
```
nauh, line 111: typo: "cehck"

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
```diff
@@ -0,0 +1,226 @@
...
+    print(credentials.valid)  # logger.debugs True
+    auth_token = credentials.token
+    print(auth_token)
```
nauh, line 114: remove "print()" statements, if they are for debugging only

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
```diff
@@ -0,0 +1,226 @@
...
+from airflow.operators.bash import BashOperator
+from airflow.operators.python import BranchPythonOperator
+
+from airflow.providers.google.cloud.operators.dataplex import (
+    DataplexCreateTaskOperator,
+    DataplexDeleteTaskOperator,
+    DataplexGetTaskOperator,
+    DataplexListTasksOperator,
+)
+from airflow.providers.google.cloud.sensors.dataplex import DataplexTaskStateSensor
```
nauh, line 27: Remove this? I don't see where it's used.

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
```diff
@@ -0,0 +1,226 @@
...
+
+from airflow import models
+from airflow.operators.bash import BashOperator
+from airflow.operators.python import BranchPythonOperator
+
+from airflow.providers.google.cloud.operators.dataplex import (
+    DataplexCreateTaskOperator,
+    DataplexDeleteTaskOperator,
+    DataplexGetTaskOperator,
+    DataplexListTasksOperator,
```
nauh, line 25: Remove these lines? I don't see where they are used.

nauh (approved):

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def _get_dataplex_job_state() -> str:
+    """
+    This method will try to get the status of the job till it is in either 'SUCCEEDED' or 'FAILED' state.
+    Args:
+    Returns: str
+    """
+    task_status = get_clouddq_task_status()
+    while (task_status != 'SUCCEEDED' and task_status != 'FAILED'):
```
nauh, line 161: This should check for all terminal states. For example, if the status is ABORTED, then this will be infinite loop.

nauh, line 161: This should check for all terminal states. For example, if the status is ABORTED, then this will be infinite loop.

nauh (changes requested):

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
AmandeepSinghCS: updated

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
AmandeepSinghCS: updated

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
AmandeepSinghCS: yes it is not used now. so removed this

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
AmandeepSinghCS: removed the unused operators

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
AmandeepSinghCS: Cancelled and Aborted are added in the updated release. Earlier these were not available. I have updated the code for composer script. We also need to update the dataplex_client in clouddq with these changes. Please create a task for this in buganizer so we don't miss it.

nauh: Got it. Actually thinking about this a little more, is it safer to check if the task_status is one of the known non-terminal states? In the worst case, this code could falsely tell user their job fails (if a new non-terminal state is added), which still seems better than running the loop infinitely IMO.

nauh (approved):

PR #179, opened 2022-07-30
Update custom sql rule failed records logic

AmandeepSinghCS:


cloudbuild.yaml
```diff
@@ -10,8 +10,6 @@ steps:
         make test test_dataplex_metadata
         make test test_dataplex_metadata_uri_templates
         make test test_dataplex_dq_configs_cache
-        make test test_dataplex_integration_performance
-        make test test_dataplex_task
```
nauh, line 14: Why do we remove these two tests?

nauh, line 14: Why do we remove these two tests?

AmandeepSinghCS: I will be adding them eventually, cloud build was giving timeout issues so removed them for now. I will create the separate PR for the same.

cloudbuild.yaml
```diff
@@ -30,5 +30,5 @@ steps:
     - 'DATAPLEX_TASK_SA=${_DATAPLEX_TASK_SA}'
 timeout: 3600s
 options:
-  machineType: 'E2_HIGHCPU_8'
```
nauh, line 33: Can we make this change in a separate/independent PR?

docs/examples/advanced_rules/accuracy_distribution_based.yaml
```diff
@@ -201,30 +196,9 @@ rule_bindings:
     entity_id: ACCURACY_CHECK_DISTRIBUTION_NOT_OK
     column_id: FULL_NAME
     row_filter_id: NONE
-    reference_columns_id: ACCURACY_CHECK_DISTRIBUTION_NOT_OK_REFERENCE_COLUMNS
```
nauh, line 204: Why do we have all these "reference_columns" changes in a PR which is about supporting "failed records for custom SQL rule"? If not directly related, can we make the changes in separate PRs?

AmandeepSinghCS: 1. I have removed reference_columns from rule_bindings, as this is optional so wanted to have few tests without specifying them.
2. Secondly, as per logic change we need to select the columns in custom_sql_statement that we are using in reference_columns. In that case we will have to update rule definition.

cloudbuild.yaml
AmandeepSinghCS: I will revert the change, this was for fixing the intermittent timeout issue.

cloudbuild.yaml
AmandeepSinghCS: The tests have been included back, as the timeout and the batches issue has been fixed after the rollback from dataproc team.

docs/examples/advanced_rules/accuracy_distribution_based.yaml
nauh: Got it. Can we please separate the changes into two PRs then? One is for optional reference_columns. The other is for custom_sql_statement.

AmandeepSinghCS: If I include, reference columns then I have to update the rules, to select the columns used in "**reference_columns**" or select all columns in **custom_sql_statement**. So this change is for **custom_statement** itself which will also cover the optional **reference_columns** scenario.

nauh (approved):

PR #181, opened 2022-08-09
Save failed records query per rule

AmandeepSinghCS:


clouddq/lib.py
```diff
@@ -131,13 +131,27 @@ def create_rule_binding_view_model(
...
+        failed_records_sql_string = failed_records_template.render(configs)
+        existing_configs = configs.get("configs")
+        existing_configs[
+            f"{rule_binding_id}_{rule_id}_failed_records_sql_string"
+        ] = failed_records_sql_string
+        configs.update({"configs": existing_configs})
+        sql_string = template.render(configs)
+        generated_sql_string_dict[
+            f"{rule_binding_id}_generated_sql_string"
+        ] = sql_string
```
nauh, line 148: Should these lines be outside of for-loop? Otherwise, we are just keep overriding `generated_sql_string_dict[f"{rule_binding_id}_generated_sql_string"]`?

clouddq/templates/dbt/macros/macros.sql
```diff
@@ -53,7 +55,91 @@
...
+      FROM (
+      {{ rule_configs.get("rule_sql_expr") }}
+      ) custom_sql
+    ) custom_sql_statement_validation_errors
+  ON
+    zero_record.rule_binding_id = custom_sql_statement_validation_errors._rule_binding_id
+{% endmacro -%}
+
+
+{% macro validate_simple_rule_failed_records_query(rule_id, rule_configs, rule_binding_id, column_name, fully_qualified_table_name, include_reference_columns) -%}
```
nauh, line 101: So..validate_simple_rule_failed_records_query and validate_complex_rule_failed_records_query do not have "failed_records_query" as part of SELECT, while validate_simple_rule and validate_complex_rule do?

When should "validate_simple_rule_failed_records_query" be used? And when "validate_simple_rule" should be used?

AmandeepSinghCS: The "**validate_simple_rule_failed_records_query**" is used in **failed_records_query** template and the "**validate_simple_rule**" will be used in **rule_binding_template**. The failed_records_query template should not have "failed_records_query" field. This template will be added as a field in rule_binding_per_rule.

nauh (approved):

PR #187, opened 2022-09-06
removed execution_ts from failed_records sql query

AmandeepSinghCS:


thinhha:
Thanks for this @AmandeepSinghCS . I want to confirm that this removes the execution_ts ONLY from the failed_records query? I want to make sure this does not lead to any backwards breaking change to existing queries used to create dq_summary.

AmandeepSinghCS:
> Thanks for this @AmandeepSinghCS . I want to confirm that this removes the execution_ts ONLY from the failed_records query? I want to make sure this does not lead to any backwards breaking change to existing queries used to create dq_summary.

Yes, @thinhha this will only remove execution_ts from the failed_records_query. The dq_summary and the target table will still have the execution_ts.

thinhha:
OK thanks for confirming. Could you let me know why execution_ts is also removed from most of the test templates? aren't they also needed for the standard tests for the sql generated by dq_summary.

thinhha (approved):
LGTM -thanks!

AmandeepSinghCS:
> OK thanks for confirming. Could you let me know why execution_ts is also removed from most of the test templates? aren't they also needed for the standard tests for the sql generated by dq_summary.

@thinhha the execution_ts is removed from failed_records_query field in the sql generated, but not from the actual generated sql that contains failed_records_query. This is a subfield inside the generated sql, which has been modified.

nauh (approved):

PR #190, opened 2022-09-19
removed reference columns from custom sql statement rule

AmandeepSinghCS:


clouddq/templates/dbt/macros/macros.sql
```diff
@@ -20,9 +20,6 @@
     '{{ fully_qualified_table_name }}' AS table_id,
     '{{ column_name }}' AS column_id,
     data.{{ column_name }} AS column_value,
-    {% for ref_column_name in include_reference_columns %}
-        data.{{ ref_column_name }} AS {{ ref_column_name }},
-    {%- endfor -%}
```
nauh, line 25: Are changes to simple rule validation intended?

clouddq/templates/dbt/macros/create_rule_binding_view.sql
```diff
@@ -107,12 +107,6 @@ all_validation_results AS (
     r.simple_rule_row_is_valid AS simple_rule_row_is_valid,
     r.complex_rule_validation_errors_count AS complex_rule_validation_errors_count,
     r.complex_rule_validation_success_flag AS complex_rule_validation_success_flag,
-    {% for ref_column_name in include_reference_columns %}
-        r.{{ ref_column_name }} AS {{ ref_column_name }},
-        {% if loop.last %}
-            {{ '\n' }}
-        {% endif %}
-    {%- endfor -%}
```
nauh, line 115: Does this affect simple rule validation?

clouddq/templates/dbt/macros/macros.sql
AmandeepSinghCS: Yes, because when both simple rule and complex rule run in the same rule_binding. The rule_binding sql will have different number of columns in the union that will cause the execution to fail.

clouddq/templates/dbt/macros/create_rule_binding_view.sql
AmandeepSinghCS: no. because this is for temporary rule_binding view that we create. reference columns are not needed in here. We have added earlier as the behavior was same at rule_binding_level, but now the behavior is changed to rule_id level. This will not impact the failed records query sql.

nauh (approved):

thinhha (approved):

PR #192, opened 2022-09-21
fix the dataplex entity for multiple matches for bigquery uri

AmandeepSinghCS:


clouddq/classes/dq_configs_cache.py
```diff
@@ -534,7 +534,7 @@ def is_dataplex_entity(
             data_path=entity_uri.get_entity_id(),
         )
         logger.info(f"Retrieved Dataplex Entities:\n{pformat(dataplex_entities_match)}")
-        if len(dataplex_entities_match) != 1:
+        if len(dataplex_entities_match) == 0:
```
nauh, line 537: Line 548: "dataplex_entity = dataplex_entities_match[0]"

If we expect the user to have the full/exact data_path (e.g. bigquery://) in `entity_uri` when they define the rule, shouldn't we expect 0 or exactly 1 entity to be found? Wondering why we expect >1 matched entities...

thinhha: You are right this should be changed to expect exact matches and have different error handling branches for 0 records found vs multiple records found.

thinhha:
To address feedback from @nauh, we should
1) Change line 310 in https://github.com/GoogleCloudPlatform/cloud-data-quality/blob/main/clouddq/integration/dataplex/clouddq_dataplex.py#L310 to remove `starts_with` so we should always expect `len(dataplex_entities_match)` to contain either 0 or 1 result from the query. 
2) revert line 537 in https://github.com/GoogleCloudPlatform/cloud-data-quality/blob/main/clouddq/classes/dq_configs_cache.py#L537 back to `if len(dataplex_entities_match) != 1: `
3) Add a new error checking branch on `if len(dataplex_entities_match) > 1: ` and log a warning message saying `ValueError: Unexpected number of Dataplex entities found for entity_uri ${entity_uri}:\n${pformat(dataplex_entities_match)}. Only one match is expected.` and return False
4) to stop the bug in line 542 of https://github.com/GoogleCloudPlatform/cloud-data-quality/blob/main/clouddq/classes/dq_configs_cache.py#L542, we need to change `f"{pformat(json.dumps(dataplex_entities_match))}\n\n"` to `f"{pformat(json.dumps(dataplex_entities_match.to_dict()))}\n\n"`

nauh (dismissed):

clouddq/classes/dq_configs_cache.py
```diff
@@ -544,6 +544,13 @@ def is_dataplex_entity(
                 f"{pformat(entity_uri.to_dict())}\n\n"
             )
             return False
+        if len(dataplex_entities_match) > 1:
+            logger.warning(
+                "ValueError: Unexpected number of Dataplex entities found for entity_uri "
+                f"${entity_uri}:\n${pformat(dataplex_entities_match)}. "
```
thinhha, line 550: `entity_uri` should be `entity_uri.complete_uri_string`

clouddq/classes/dq_configs_cache.py
```diff
@@ -544,6 +544,13 @@ def is_dataplex_entity(
                 f"{pformat(entity_uri.to_dict())}\n\n"
             )
             return False
+        if len(dataplex_entities_match) > 1:
+            logger.warning(
+                "ValueError: Unexpected number of Dataplex entities found for entity_uri "
```
thinhha, line 549: As this is not a Python error but a warning, let's remove `ValueError: `

thinhha (approved):
Thanks @AmandeepSinghCS - LGTM!

PR #193, opened 2022-09-26
Include star for ref columns

AmandeepSinghCS:


clouddq/classes/dq_rule_binding.py
```diff
@@ -348,6 +359,7 @@ def resolve_all_configs_to_dict(
                     "column_id": self.column_id,
                     "reference_columns_id": self.reference_columns_id,
                     "include_reference_columns": include_reference_columns,
+                    "include_all_reference_columns": include_all_reference_columns,
```
nauh, line 362: Where is this used?

clouddq/classes/dq_rule_binding.py
```diff
@@ -334,11 +336,20 @@ def resolve_all_configs_to_dict(
             row_filter_config = self.resolve_row_filter_config(configs_cache)
             # resolve reference columns config
             if self.reference_columns_id:
+                include_all_reference_columns = False
                 include_reference_columns = self.resolve_reference_columns_config(
                     configs_cache
                 ).include_reference_columns
+                if include_reference_columns[0] == "*":
```
nauh, line 343: Can we also check if the length of the include_reference_columns is 1?

clouddq/classes/dq_rule_binding.py
AmandeepSinghCS: @nauh I have added this in configs dictionary to identify the configs for job if it required all columns or no with this flag.

clouddq/classes/dq_rule_binding.py
AmandeepSinghCS: in case we have both * and few other columns in reference columns section, I should raise an error or log a warning and continue with execution with including reference columns.

nauh: I think it's probably better to raise an error if the list include "*" and other columns.

clouddq/classes/dq_rule_binding.py
```diff
@@ -334,11 +336,20 @@ def resolve_all_configs_to_dict(
...
+                    column_names = bigquery_client.get_table_columns(
+                        table=table_entity.get_table_name(),
+                        project_id=table_entity.instance_name,
+                    )
```
thinhha, line 347: what kind of errors does `bigquery_client.get_table_columns()` throw and where are they being handled? e.g. what happens when `table_entity` doesn't exists, when connection fails, or when the current user does have sufficient authentication to connect? Are we assuming that if the code reached this point we shouldn't get those errors?

clouddq/classes/dq_rule_binding.py
thinhha: agree we should check and throw error if user include "*" and other columns

clouddq/classes/dq_rule_binding.py
thinhha: if any of the above errors happen, will the expected behaviour be for the client itself to throw and error and exit, i.e. we don't need to do anything else here?

tests/resources/configs/reference_columns/reference-columns.yml
```diff
@@ -26,3 +26,11 @@ reference_columns:
...
       - source
       - status
+
+  CONTACT_DETAILS_ALL_REFERENCE_COLUMNS:
+    include_reference_columns:
+      - "*"
+
+  311_SERVICE_REQUESTS_ALL_REFERENCE_COLUMNS:
+    include_reference_columns:
+      - "*"
```
thinhha, line 36: should both of these reference IDs just be a single one called "INCLUDE_ALL_REFERENCE_COLUMNS"?

clouddq/classes/dq_rule_binding.py
AmandeepSinghCS: yes the exception handling is done in the bigquery_client if the table entity doesn't exists or unable to access table entity.

nauh (dismissed):

tests/resources/configs/reference_columns/reference-columns.yml
AmandeepSinghCS: incorporated

clouddq/classes/dq_rule_binding.py
```diff
@@ -334,11 +336,32 @@ def resolve_all_configs_to_dict(
             row_filter_config = self.resolve_row_filter_config(configs_cache)
             # resolve reference columns config
             if self.reference_columns_id:
+                include_all_reference_columns = False
                 include_reference_columns = self.resolve_reference_columns_config(
                     configs_cache
                 ).include_reference_columns
```
thinhha, line 342: this is obsolete if "*" is in include_reference_columns

thinhha: could we optimize the if condition clauses a bit more? there are duplicated and redundant work still happening

clouddq/classes/dq_rule_binding.py
```diff
@@ -334,11 +336,32 @@ def resolve_all_configs_to_dict(
...
+                if (
+                    "*" in include_reference_columns
```
thinhha, line 344: this is duplicated

nauh (approved):

PR #210, opened 2022-11-21
Adding Reference Column to view

shourya116:


AmandeepSinghCS:
@shourya116 Adding reference columns to the views will create issues, if we are combining simple and complex rule in single rule binding. As the number of columns will be different in case of them that will cause UNION ALL of queries to fail and hence the execution will fail. We can discuss on possible solutions with team.

PR #213, opened 2022-12-15
fixed complex_rule_validation_success_flag

shourya116:


configs/reference_columns/reference-columns.yml
```diff
@@ -22,4 +22,4 @@ reference_columns:
 
   INCLUDE_ALL_REFERENCE_COLUMNS:
     include_reference_columns:
-      - *
+      - "*"
```
AmandeepSinghCS, line 25: @shourya116 create a separate PR for this change.

PR #228, opened 2023-02-22
removed requests_oauth2 dependency from composer DAG

shourya116:


AmandeepSinghCS (approved):
LGTM, Thank you!!

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
```diff
@@ -91,7 +91,7 @@
 }
 
 
-def get_session_headers() -> tuple:
+def get_session_headers() -> dict:
```
AmandeepSinghCS, line 94: Please change the method documentation, to return dictionary then tuple on L98

thinhha:
Hi @shourya116 this looks good to me. Could you please raise a separate PR to change the DAG such that the delete step is always done at the end?

PR #229, opened 2023-02-28
updated the composer dag

shourya116:


scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
```diff
@@ -156,6 +156,20 @@ def _get_dataplex_job_state() -> str:
...
+    Args:
+    Returns: str
+    """
+    headers = get_session_headers()
+    res = requests.get(
+        f"{DATAPLEX_ENDPOINT}/v1/projects/{DATAPLEX_PROJECT_ID}/locations/{DATAPLEX_REGION}/lakes/{DATAPLEX_LAKE_ID}/tasks/{DATAPLEX_TASK_ID}",
+        headers=headers)
+    if res.status_code == 200:
+        return "task_exist"
+    else:
```
thinhha, line 171: can we instead please check res.status_code == 404 for task not exist, and in the "else" branch return an error?

thinhha (changes requested):
Thanks @shourya116. Could you please take a look at the HTTP response semantics and only return the "task_not_exist" code for the 404 response? We need a separate branch to raise an error in case the response is neither 200 or 404.

scripts/dataproc-workflow-composer/clouddq_composer_dataplex_task_job.py
shourya116: I have incorporated the suggestion.

thinhha (approved):
Lgtm thanks Shourya

PR #230, opened 2023-02-28
fixed cloud build issue for pytz

AmandeepSinghCS:


thinhha:
Could we try updating the requirements.txt file instead of patching the test script?

thinhha (changes requested):
We should not patch a test script. Could we try adding pytz to the requirements.txt file?

PR #235, opened 2023-03-30
fixed Dataplex URI for Big Lake Tables

shourya116:


thinhha (approved):
Thanks Shourya. Looks good to me.

PR #244, opened 2023-06-23
adding metadata to logs

shourya116:


clouddq/log.py
```diff
@@ -64,14 +64,15 @@ def format(self, record):
         return super().format(record)
 
 
-def add_cloud_logging_handler(logger: Logger):
+def add_cloud_logging_handler(metadata, logger: Logger):
```
thinhha, line 67: please add type hint, also new arguments should be added *after* existing arguments, rather than before

clouddq/log.py
```diff
@@ -64,14 +64,15 @@ def format(self, record):
...
     client = google.cloud.logging.Client()
     handler = CloudLoggingHandler(
         client=client,
         name="clouddq",
         labels={
             "name": APP_NAME,
             "releaseId": APP_VERSION,
+            "metadata": str(metadata),
```
thinhha, line 75: Please ensure the metadata is not included as a string but as a nested object in the logs. I think this means instead of `str()` we need to use `json.dumps()`

clouddq/main.py
```diff
@@ -270,8 +278,26 @@ def main(  # noqa: C901
...
+        metadata = json.loads(metadata)
         # Set-up cloud logging
-        add_cloud_logging_handler(logger=json_logger)
+        if add_metadata_to_logs:
+            dataplex_batch_name = os.environ.get("DATAPLEX_BATCH_NAME")
+            dataplex_task_name = os.environ.get("DATAPLEX_TASK_NAME")
+            if dataplex_batch_name:
+                metadata["dataplex_batch_name"] = dataplex_batch_name
+            if dataplex_task_name:
+                metadata["dataplex_task_name"] = dataplex_task_name
```
thinhha, line 290: can we move this out of the ifelse

clouddq/main.py
```diff
@@ -270,8 +278,26 @@ def main(  # noqa: C901
...
+            add_cloud_logging_handler(logger=json_logger, metadata=metadata)
+        else:
+            dataplex_batch_name = os.environ.get("DATAPLEX_BATCH_NAME")
+            dataplex_task_name = os.environ.get("DATAPLEX_TASK_NAME")
+            if dataplex_batch_name:
+                metadata["dataplex_batch_name"] = dataplex_batch_name
+            if dataplex_task_name:
+                metadata["dataplex_task_name"] = dataplex_task_name
+            add_cloud_logging_handler(logger=json_logger, metadata=metadata)
```
thinhha, line 299: if not add_metadata_to_logs: should we not exclude the "metadata" dictionary and only include the dataplex_batch_name and dataplex_task_name?

clouddq/main.py
shourya116: If we move this out than both the variables will be included into the metadata dictionary. Instead  if add_metadata_to_logs is set to false we need to create a new dictionary for both the variables and pass it into the add_cloud_logging_handler().

shourya116: we can surely move out the part where we are reading the values for the variables.

